PR #107, opened 2015-05-01
Sly typos

abelards:
Hi guys, so far it's been super easy to use even if I don't have much sysadm-fu.
I'm picky about typos, grammar and english so here you are.
It's great being able to contribute even to these tiny things ;)


Soulou:
That's great, thanks a lot.

PR #111, opened 2015-05-05
Add helpers for each addon in order to access their interactive console

Soulou:


db/mongo_console.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	opts := apps.RunOpts{
+		App: app,
+		Cmd: []string{"mongo", "-u", user, "-p", password, mongoURL.Host + "/" + user},
+	}
+
+	err = apps.Run(opts)
+	if err != nil {
+		return errgo.Newf("Fail to run redis console: %v", err)
```
q42jaap: error message should probably mention "mongo" not "redis"

Soulou: Hi there, you're completely right, and it has been fixed there: https://github.com/Scalingo/cli/commit/26b9d8f20a5d013cc9e1305bbb8df40eff1419c8

Fixed in next release

PR #143, opened 2015-09-21
Fix #137 : collaborators-add completion now uses goroutines

### LINKED ISSUE #137: Autocomplete: collaborators-add should be less resource-expensive
When you want to complete [collaborators-add](https://github.com/Scalingo/cli/blob/master/cmd/autocomplete/collaborators_add.go) command, it will show you every collaborator you are working with, from any of your other apps.
If you manage to have a lot of apps, the client itself will make the collaborators-list request to Scalingo API for EACH app, which will take time.

An improvement would be to do asynchronous requests (even if it's hard in its current implementation), or do a new server-side function, maybe?

Zyko0:
When you are completing by doing `<tab><tab>`, `collaborators-add` can have a lot of requests to do depending on you apps' number.

### DIFF
--- a/cmd/autocomplete/collaborators_add.go
+++ b/cmd/autocomplete/collaborators_add.go
@@ -54,6 +54,7 @@ func CollaboratorsAddAutoComplete(c *cli.Context) error {
 		}
 	}()
 	wg.Wait()
+	close(ch)
 
 	for email, _ := range setEmails {
 		isAlreadyCollaborator := false
--- a/cmd/autocomplete/collaborators_add.go
+++ b/cmd/autocomplete/collaborators_add.go
@@ -27,11 +27,11 @@ func CollaboratorsAddAutoComplete(c *cli.Context) error {
 		return nil
 	}
 
+	ch := make(chan string)
 	var wg sync.WaitGroup
 	wg.Add(len(apps))
-	setEmails := make(map[string]bool)
 	for _, app := range apps {
-		go func(app *api.App, setEmails map[string]bool) {
+		go func(app *api.App) {
 			defer wg.Done()
 			appCollaborators, erro := api.CollaboratorsList(app.Name)
 			if erro != nil {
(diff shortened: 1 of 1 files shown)


cmd/autocomplete/collaborators_add.go
```diff
@@ -25,15 +27,25 @@ func CollaboratorsAddAutoComplete(c *cli.Context) error {
...
+		go func(app *api.App, setEmails map[string]bool) {
+			appCollaborators, erro := api.CollaboratorsList(app.Name)
+			if erro != nil {
+				err = erro
+				return
+			}
+			for _, col := range appCollaborators {
+				setEmails[col.Email] = true
+			}
+			wg.Done()
```
Soulou: User `defer wg.Done()` like that it will be called in all cases

cmd/autocomplete/collaborators_add.go
```diff
@@ -25,15 +27,25 @@ func CollaboratorsAddAutoComplete(c *cli.Context) error {
...
-			setEmails[col.Email] = true
-		}
+		go func(app *api.App, setEmails map[string]bool) {
+			appCollaborators, erro := api.CollaboratorsList(app.Name)
+			if erro != nil {
+				err = erro
+				return
+			}
+			for _, col := range appCollaborators {
+				setEmails[col.Email] = true
```
Soulou: map in go are not "concurrency safe", you need to set a mutex to lock it or to use channels (which the the recommanded way)

cmd/autocomplete/collaborators_add.go
```diff
@@ -25,15 +27,25 @@ func CollaboratorsAddAutoComplete(c *cli.Context) error {
...
-		if err != nil {
-			return nil
-		}
-		for _, col := range appCollaborators {
-			setEmails[col.Email] = true
-		}
+		go func(app *api.App, setEmails map[string]bool) {
+			appCollaborators, erro := api.CollaboratorsList(app.Name)
+			if erro != nil {
+				err = erro
```
Soulou: In this case we are looking for completion, if there are error, maybe logging them directly is enough?

Zyko0: Do you mean that we should remove `erro` temporary variable?
If we only log the error, how do we handle a goroutine fail?

Soulou: Just log the error and return

cmd/autocomplete/collaborators_add.go
```diff
@@ -25,16 +27,34 @@ func CollaboratorsAddAutoComplete(c *cli.Context) error {
...
+				return
+			}
+			for _, col := range appCollaborators {
+				ch <- col.Email
+			}
+		}(app)
 	}
+	if err != nil {
```
Soulou: Checking the error ere has no real meaning, it should be done after the go routines have ended

PR #163, opened 2015-10-19
Fix #151: Destroy command is always run

### LINKED ISSUE #151: destroy command is always run
The command `scalingo -a invalidapp destroy` is always asking you to confirm the app you want to destroy even if the app doesn't exist.
We should maybe check if the app is existing before asking the user to re-type its app name.

Zyko0:
`scalingo destroy` was always run, whatever appname you gave.
I thought it would be better to check if the app is existing or not first.

So two requests instead of one, but the user has to type its app name again to confirm.
If he has to confirm an app that doesn't exists, it's even more bad than a 100ms request.

### DIFF
--- a/apps/destroy.go
+++ b/apps/destroy.go
@@ -11,7 +11,7 @@ import (
 func Destroy(appName string) error {
 	var validationName string
 
-	res, err := scalingo.AppsShow(appName)
+	_, err := scalingo.AppsShow(appName)
 	if err != nil {
 		return errgo.Mask(err, errgo.Any)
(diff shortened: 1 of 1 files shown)


apps/destroy.go
```diff
@@ -10,13 +10,19 @@ import (
 
 func Destroy(appName string) error {
 	var validationName string
+
+	res, err := scalingo.AppsShow(appName)
```
Soulou: res is not used, please use `_`

PR #202, opened 2016-03-02
Remove redundant condition in isAddrInUse

tevino:


Soulou:
Thanks for the fix :)

PR #211, opened 2016-05-11
Add short option `-s` for `--synchronous` in `scale` and `restart` commands

cbliard:


cmd/restart.go
```diff
@@ -23,7 +23,7 @@ var (
 	  scalingo --app my-app restart web-1`,
 		Action: func(c *cli.Context) {
 			currentApp := appdetect.CurrentApp(c)
-			if err := apps.Restart(currentApp, c.Bool("synchronous"), c.Args()); err != nil {
+			if err := apps.Restart(currentApp, c.Bool("synchronous") || c.Bool("s"), c.Args()); err != nil {
```
Soulou: Not necessary, both forms can't be present in the command line, and the same value is set for the long and short flag.

cmd/scale.go
```diff
@@ -24,7 +24,7 @@ var (
 			currentApp := appdetect.CurrentApp(c)
 			if len(c.Args()) == 0 {
 				cli.ShowCommandHelp(c, "scale")
-			} else if err := apps.Scale(currentApp, c.Bool("synchronous"), c.Args()); err != nil {
+			} else if err := apps.Scale(currentApp, c.Bool("synchronous") || c.Bool("s"), c.Args()); err != nil {
```
Soulou: Same as before comment

PR #227, opened 2016-06-08
Better typo if a notification is already enabled

utay:


notifications/provision.go
```diff
@@ -30,3 +31,11 @@ func Provision(app, webHookURL string) error {
 	}
 	return nil
 }
+
+func getType(webHookURL string) string {
```
Soulou: I don't like that, the type should be saved in the notification, and read directly from the struct, no logics like that should appear on the client side

Soulou: Actually it should be read from the error returned by the server.

notifications/provision.go
```diff
@@ -18,7 +19,7 @@ func Provision(app, webHookURL string) error {
 	c := config.ScalingoClient()
 	params, err := c.NotificationProvision(app, webHookURL)
 	if err != nil {
-		return errors.New("Notification " + webHookURL + " can't be enabled because you have already an other notification of the same type enabled!")
+		return errors.New("Notification to " + webHookURL + " can't be enabled because you have already an other notification of the same type " + getType(webHookURL) + " enabled!")
```
Soulou: You should use standard validation error handling which reads the error from the server

PR #236, opened 2016-06-23
Feature/add size flag

### LINKED ISSUE #233: [One-off] User should be able to choose the size of the container
For instance:

```
scalingo -a appname run --size L bash
```

Things to do:

- Handle the -s/--size flag
- Send its value to the API

EtienneM:
Fix #233 

Work in progress

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -1,10 +1,10 @@
 package cmd
 
 import (
-	"github.com/Scalingo/codegangsta-cli"
 	"github.com/Scalingo/cli/appdetect"
 	"github.com/Scalingo/cli/apps"
 	"github.com/Scalingo/cli/cmd/autocomplete"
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -64,6 +71,7 @@ var (
 			opts := apps.RunOpts{
 				App:    currentApp,
 				Cmd:    c.Args(),
+				Size:   c.String("z"),
```
Soulou: Why `z` ?

EtienneM: `s` is already for the silent flag...

Soulou: I would remove the -s flag for --silent and use it for size, it's more.. common

EtienneM: I just noticed this comment... I will modify that.

Soulou:
I did it, its ok

PR #239, opened 2016-06-24
Add auto key detection

johnsudaar:


Soulou:
Wasn't there the idea of "if there is no key, create one"?

Soulou:
And additionaly, there should be no logic in the cmd package other than flag logic, please export it somewhere else

Soulou:
I don't like the idea of the --auto flag, it should be implemented differently

Issue #267, opened 2016-12-07
Clarify domains-ssl signature


premasagar:
In an earlier version of the CLI, `domains-ssl` accepted the domain name before the `--cert` / `--key` arguments. The help text uses that in its example:

```
$ scalingo domains-ssl --help

> $ scalingo -a myapp domains-ssl example.com --cert <certificate.crt> --key <keyfile.key>
```

But that doesn't work. I get a non-descript error message:

```
> No help topic for 'domains-ssl'
```

Moving the domain name to the end of the command works:

```
> $ scalingo -a myapp domains-ssl --cert <certificate.crt> --key <keyfile.key> example.com
```

PR #290, opened 2017-06-01
Update to original cli lib

bragagia:
Use urfave/cli instead of Scalingo/codegangsta-cli which hasn't been updated since a long time.


bragagia:
It solves #267

bragagia:
I think [this repo](https://github.com/Scalingo/codegangsta-cli) if it is not used elsewhere.

PR #292, opened 2017-06-01
Adds timeout on update check

bragagia:
It solves #274

### DIFF
--- a/update/check.go
+++ b/update/check.go
@@ -42,7 +42,6 @@ func Check() error {
 	<-gotLastVersion
 
 	if gotAnError {
-		fmt.Println("Update checker: connection error")
 		return errgo.New("Update checker: connection error")
 	}
 	if version == lastVersion {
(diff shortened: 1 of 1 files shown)


Soulou:
Disagreed, please add a timeout in the method `getLastVersion`

bragagia:
If so, we need to create new goroutine inside `getLastVersion`, is that really useful ?

Soulou:
No need to add a goroutine, you can set a timeout on the HTTP client :-)

bragagia:
Done :)

update/check.go
```diff
@@ -37,6 +40,11 @@ func Check() error {
 	}
 
 	<-gotLastVersion
+
+	if gotAnError {
+		fmt.Println("Timeout when connecting on scalingo server.")
```
Soulou: Why?

bragagia: I did'nt noticed that this was called on every command. Removed ;)

Issue #280, opened 2017-04-05
Update message should be displayed in red


yannski:
"-----> Your Scalingo client (1.4.1) is out-of-date: some features may not work correctly.
       Please update to '1.5.0' by reinstalling it: http://cli.scalingo.com
"

EtienneM:
Fixed

PR #297, opened 2017-06-20
Update information line is now red, Fixing #280

MrMourse:


Soulou:
Actually, we're already using colors in the program, and here you're using a totaly new approach to do that, could you use what is already existing?

cmd/addons_providers_plans_list.go
```diff
@@ -14,7 +14,7 @@ var (
 		Usage:       "List plans",
 		Action: func(c *cli.Context) {
 			if len(c.Args()) != 1 {
-				cli.ShowCommandHelp(c, "addon-plans")
+				cli.ShowCommandHelp(c, "addons-plans")
```
johnsudaar: Why is this in this pr?

PR #301, opened 2017-06-21
Notification v2

MrYawe:
Notifier:
- [x] Create command
- [x] Update command
- [x] Delete command
- [x] List command
- [x] Show details command

Notification platform:
- [x] List command


MrYawe:
@Soulou Should we delete notification command ?

cmd/notification_platforms.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/Scalingo/cli/cmd/autocomplete"
+	"github.com/Scalingo/cli/notification_platforms"
+	cli "github.com/Scalingo/codegangsta-cli"
+)
+
+var (
+	NotificationPlatformListCommand = cli.Command{
+		Name:        "platforms-list",
```
Soulou: I'm not super fan of the command name, `notification-platforms` would be better no?

Soulou: I made the choice to transform it to `notification-platforms` as stated by the variable name ;)

PR #313, opened 2017-09-13
Connect with tls when available

johnsudaar:


johnsudaar:
Needs patch and migration from https://github.com/Scalingo/appsdeck-database/pull/408

PR #327, opened 2018-01-12
Send tty size on first container byte

### LINKED ISSUE #326: TTY Size is sent too early on run commands

johnsudaar:
Fix #326


Soulou:
Tested and worked for me

PR #328, opened 2018-01-26
Update to the new auth workflow

Soulou:


apps/run.go
```diff
@@ -254,7 +254,13 @@ func (ctx *runContext) exitCode() (int, error) {
...
 		return -1, errgo.Mask(err, errgo.Any)
 	}
-	req.SetBasicAuth("", config.AuthenticatedUser.AuthenticationToken)
+
+	token, err := config.C.TokenGenerator.GetAccessToken()
+	if err != nil {
+		return -1, errgo.Notef(err, "fail to generate auth")
+	}
+
+	req.SetBasicAuth("", token)
```
johnsudaar: Does this work?
The JWT are use with Bearer auth not Basic auth.

apps/run.go
```diff
@@ -297,7 +303,12 @@ func (ctx *runContext) connectToRunServer() (*http.Response, net.Conn, error) {
 	if err != nil {
 		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
-	req.SetBasicAuth("", config.AuthenticatedUser.AuthenticationToken)
+	token, err := config.C.TokenGenerator.GetAccessToken()
+
+	if err != nil {
+		return nil, nil, errgo.Notef(err, "fail to generate auth")
+	}
+	req.SetBasicAuth("", token)
```
johnsudaar: Same here

config/auth.go
```diff
@@ -29,67 +28,72 @@ var (
...
+				return nil, nil, errors.New("canceled by user")
 			} else {
-
 				appio.Errorf("Fail to login (%d/3): %v\n\n", i+1, err)
 			}
 		}
 	}
 	if err == ErrAuthenticationFailed {
-		return nil, errors.New("Forgot your password? https://my.scalingo.com")
+		return nil, nil, errors.New("Forgot your password? https://my.scalingo.com")
```
johnsudaar: auth.scalingo.com

config/config.go
```diff
@@ -100,18 +100,21 @@ func init() {
 		TlsConfig.MinVersion = tls.VersionTLS10
 	}
 
-	AuthenticatedUser, err = Authenticator.LoadAuth()
-	if err == nil {
-		C.apiToken = AuthenticatedUser.AuthenticationToken
+	user, tokenGenerator, err := Authenticator.LoadAuth()
```
EtienneM: What happens if `err != nil`?

config/auth.go
Soulou: Done

apps/run.go
Soulou:
```
└> ./scalingo -a db run bash
-----> Connecting to container [one-off-1598]...  
-----> Process 'bash' is starting...  

   _____               __ _                    
  / ___/ _____ ____ _ / /(_)____   ____ _ ____ 
  \__ \ / ___// __ `// // // __ \ / __ `// __ \
 ___/ // /__ / /_/ // // // / / // /_/ // /_/ /
/____/ \___/ \__,_//_//_//_/ /_/ \__, / \____/ 
                                /____/         

```

Soulou: Authentication too permissive?

Soulou: Oh I know actually

Soulou: The token is sent to the run service (`one-off.scalingo.com`) which still expect basic auth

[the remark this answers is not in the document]
Soulou: Explanation above

apps/run.go
Soulou: Explanation above

config/config.go
Soulou: user and tokenGenerator will be nil, so no authentication will be loaded and the CLI will run in "unauthenticated" mode. So if a request requires auth, the CLI will ask the user to login

PR #331, opened 2018-01-26
Implement rename command Fixes #330

### LINKED ISSUE #330: Rename command should be present

Soulou:


apps/rename.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	validationName = strings.Trim(validationName, "\n")
+
+	if validationName != appName {
+		return errgo.Newf("'%s' is not '%s', aborting…\n", validationName, appName)
+	}
+
+	_, err = c.AppsRename(appName, newName)
+	if err != nil {
+		return errgo.Notef(err, "fail to name app")
```
EtienneM: fail to name app -> fail to **re**name app

cmd/rename.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/Scalingo/cli/appdetect"
+	"github.com/Scalingo/cli/apps"
+	"github.com/Scalingo/cli/cmd/autocomplete"
+	"github.com/urfave/cli"
+)
+
+var (
+	RenameCommand = cli.Command{
+		Name:     "rename",
+		Category: "Global",
```
EtienneM: Shouldn't it be `App Management`?

Soulou: True, same for `destroy`, updated in next commit

PR #338, opened 2018-03-07
Migrate to dep

EtienneM:


Soulou:
I think go-scalingo version has changed with an ancient version

EtienneM:
I updated it so that it is in sync with `master`.

PR #339, opened 2018-03-07
Add CRUD for autoscalers

EtienneM:
Please first review https://github.com/Scalingo/go-scalingo/pull/40

- [x] Add bash autocomplete stuff
- [x] `scalingo ps` command should display somehow that a given container type has an autoscaler
- [x] When manual scaling, warn about disabling the autoscaling and ask for confirmation
- [x] Add Update
- [x] Add Remove

Related to Scalingo/autoscaler-service#1


EtienneM:
```text
╰─$ scalingo -a biniou autoscalers

+----------+----------------+-------------------+-----------+----------------+---------------+
| DISABLED | CONTAINER TYPE |      METRIC       |  TARGET   | MIN CONTAINERS | MAX CONTAIERS |
+----------+----------------+-------------------+-----------+----------------+---------------+
| true     | web            | p95_response_time | 60.000000 | 2              | 7             |
+----------+----------------+-------------------+-----------+----------------+---------------+
```

EtienneM:
```text
╰─$ scalingo -a biniou autoscalers-add --container-type web --metric cpu --target 0.75 --min-containers 1 --max-containers 3

-----> Autoscaler created on biniou for web containers
```

Soulou:
I disagree on the first table column being `Disabled`. Why using a negative form instead of `ACTIVE` ? It should be the last one.

I think the autoscaler row should include the current scaling of the application, to avoid people to make 2 requests here.

Soulou:
So finally, no integration in the `scale` command then!

What happens if I try to scale my app with the CLI?

EtienneM:
I found it really unclear to integrate with the `scale` command...

If you manually scale with the CLI, I would like to show a message warning about an autoscaler being present and asking confirmation.

Soulou:
Ok for the integration with the scale command, but then the UI should be nice.

What about the command `scalingo ps`, shouldn't it display also that a container type is autoscaled?

For the confirmation how will you do that, fetch the autoscalers first to see if one exists and is active, or the API will returns something?

Soulou:
Can you share how you've updated the syle?

EtienneM:
You are right about the `scalingo ps` command. I will add something.

For the confirmation, I had the first solution in mind (fetch the autoscalers first) but both are doable.

I edited the previous messages with the updated wording of the table

Soulou:
I still don't know if Active should be the first column =\

EtienneM:
I had the idea to highlight with a different color the line where the autoscaler is disabled, but then it looked like a christmas tree..

EtienneM:
```text
$ scalingo -a biniou autoscalers-enable web
$ scalingo -a biniou autoscalers-disable web
$ scalingo -a biniou autoscalers-update --container-type web --max-containers 5

-----> Autoscaler updated on biniou for web containers
```

EtienneM:
```text
╰─$ scalingo -a biniou scale web:3

  /!\  An autoscaler is configured for web container. Manually scaling it will disable the autoscaler. Do you confirm? (y/N)

 !     An error occurred:
       You didn't confirm, aborting…
```

Typing anything but `y` or `Y` will abort the scaling operation

EtienneM:
```text
╰─$ scalingo -a biniou ps

+---------+--------+------+---------------------+
|  NAME   | AMOUNT | SIZE |       COMMAND       |
+---------+--------+------+---------------------+
| web (*) | 1      | M    | `sample-go-martini` |
| test    | 2      | M    | `sample-go-martini` |
+---------+--------+------+---------------------+
  (*) has an autoscaler defined
```

Soulou:
Please rebase

apps/ps.go
```diff
@@ -19,15 +20,31 @@ func Ps(app string) error {
 	t := tablewriter.NewWriter(os.Stdout)
 	t.SetHeader([]string{"Name", "Amount", "Size", "Command"})
 
+	hasAutoscaler := false
 	for _, ct := range processes {
+		name := ct.Name
+		_, err = autoscalers.GetFromContainerType(app, name)
```
Soulou: You need a cache here, you can't make a list request at each iteration of the loop

(same at the other place of the PR where you're doing this)

Soulou: @EtienneM You get all the autoscalers with a call in #index once, then you look into them instead of making plenty of API requests.

EtienneM: Silly me... Sorry. That's obvious to me now

Issue #121, opened 2015-07-23
A user should be able to reset it's deployment cache (in case of corruption mostly)


Soulou:

Soulou:
The command should be something like:

```
scalingo -a appname deployments-delete-cache
```

EtienneM:
Fixed by #358

PR #358, opened 2018-05-31
User is able to reset deployment cache

iAmoric:
Related to Scalingo/cli#121


Soulou:
@iAmoric Can you update `go-scalingo` ?

```
dep ensure -update github.com/Scalingo/go-scalingo
```

Thanks a lot

PR #360, opened 2018-06-05
Since multiple deployment can be run on parallel and the new event is not always fetched, ensure logs are streamed after 'scalingo deploy .jar/war

### LINKED ISSUE #359: [scalingo deploy] logs streaming doesn't always trigger

Soulou:
Fixes #359

### DIFF
--- a/deployments/follow.go
+++ b/deployments/follow.go
@@ -50,14 +50,22 @@ func Stream(opts *StreamOpts) error {
 		return errgo.Mask(err, errgo.Any)
 	}
 
-	var event deployEvent
+	// This method can focus on one given deployment and will on display events
+	// related to this deployment
 	currentDeployment := &scalingo.Deployment{
 		ID: opts.DeploymentID,
 	}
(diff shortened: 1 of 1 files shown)


deployments/follow.go
```diff
@@ -51,8 +51,12 @@ func Stream(opts *StreamOpts) error {
 	}
 
 	var event deployEvent
-	var currentDeployment *scalingo.Deployment
-	oldStatus := ""
+	currentDeployment := &scalingo.Deployment{
+		ID: opts.DeploymentID,
+	}
+	anyDeployment := currentDeployment.ID == ""
```
EtienneM: I don't really get the meaning of `anyDeployment`.
Moreover, maybe adding a bit of comments would help understand what happens here. In which context the `DeploymentID` is not provided?

Soulou: When you run the command `scalingo deployment-follow` which display every events about all the deployments of an app

PR #367, opened 2018-07-04
Correctly display help whe nthe users types an unknown command

### LINKED ISSUE #365: Some "faulty" commands do not return error or help message
```
$ scalingo -a scalingo-homepage-production bash
$
```

Soulou:
Fixes #365

### DIFF
--- a/scalingo/main.go
+++ b/scalingo/main.go
@@ -89,7 +89,8 @@ COMMANDS:{{range .VisibleCategories}}{{if .Name}}
 		 {{end}}{{end}}{{if .VisibleFlags}}
 GLOBAL OPTIONS:
    {{range $index, $option := .VisibleFlags}}{{if $index}}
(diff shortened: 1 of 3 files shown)


scalingo/main.go
```diff
@@ -89,7 +89,8 @@ COMMANDS:{{range .VisibleCategories}}{{if .Name}}
 		 {{end}}{{end}}{{if .VisibleFlags}}
 GLOBAL OPTIONS:
    {{range $index, $option := .VisibleFlags}}{{if $index}}
-   {{end}}{{$option}}{{end}}{{end}}{{if .Copyright}}
+   {{end}}{{$option}}{{end}}{{end}}
+{{if .Copyright}}
```
EtienneM: Is it valid to put the `if` statement on a new line?

Soulou: Yes we're in a template and I wanted to add a newline because the output was not great (really compacted)

PR #369, opened 2018-07-05
Consider CTRL^C the normal way to stop a logs -f, not an error

### LINKED ISSUE #355: [Logs] Deprecation line is not shown
This message:
```
-----> Your Scalingo client (1.7.0) is out-of-date: some features may not work correctly.
       Please update to '1.8.0' by reinstalling it: http://cli.scalingo.com
```

is not shown when using the command: `scalingo logs -f`

Soulou:
Fixes #355

### DIFF
--- a/apps/logs.go
+++ b/apps/logs.go
@@ -7,6 +7,8 @@ import (
 	stdio "io"
 	"io/ioutil"
 	"net/url"
+	"os"
+	"os/signal"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


apps/logs.go
```diff
@@ -137,6 +153,9 @@ func streamLogs(logsRawURL string, filter string) error {
 					time.Sleep(time.Second * 1)
 				}
 				continue
+
+			} else if strings.Contains(err.Error(), "use of closed network connect") {
```
EtienneM: Isn't it possible to find the root error and test against an error instead of a substring? I am sure you already tried but just want to be sure.

Soulou:
```
internal/poll/fd.go:var ErrNetClosing = errors.New("use of closed network connection")
```

Package internal of stdlib, not usable.

ie. official http2 lib, they are checking the string..:

```
net/http/h2_bundle.go:	if strings.Contains(str, "use of closed network connection") {
```

PR #374, opened 2018-08-13
Advise to restart the containers after setting environment

### LINKED ISSUE #373: After env-set advise to restart the app
Currently the output after a `env-set` is:

```
VAR_1 has been set to 'CONTENT VAR_1'.
```

It should also advise the user to restart the app in order for the environment variable to be taken into account. Just like the toaster does when setting variables in the web dashboard.

EtienneM:
```
$ scalingo -a biniou env-set TOTO=titi
TOTO has been set to 'titi'.
Please restart your containers:
scalingo --app biniou restart
```

Fix #373

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+### To be Released
+
+* [env-set] Advise to restart after setting the environment #373
+
 ### 1.9.0
 
 * [alerts] Add all command to CRUD alerts #346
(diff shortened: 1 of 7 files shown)


Soulou:
I would add an additional new line before the advice and would write something like: you need to restart your container for the environment changes to be used (with a better wording)

EtienneM:
'right and I also added the advice for the env unset

Soulou:
```
Restart your containers to apply these environment changes on your application:
```

? what do you think?

PR #383, opened 2018-09-10
Add routing settings

### LINKED ISSUE #344: Add routing settings to the CLI
- [ ] Force HTTPS
- [ ] Sticky session
- [ ] Canonical

EtienneM:
Fix #344

```text
$ scalingo -a biniou force-https [--enable]
-----> Force HTTPS has been enabled on biniou
```

```text
$ scalingo -a biniou force-https --disable
-----> Force HTTPS has been disabled on biniou
```

```text
$ scalingo -a biniou sticky-session [--enable]
-----> Sticky session has been enabled on biniou
```

```text
$ scalingo -a biniou sticky-session --disable
-----> Sticky session has been disabled on biniou
```

```text
$ scalingo -a biniou domains
+-----------------+-----+
|     DOMAIN      | SSL |
+-----------------+-----+
| biniou.fr       |   - |
| example.com (*) |   - |
+-----------------+-----+
  (*) canonical domain
```

```text
$ scalingo -a biniou set-canonical-domain biniou.fr
-----> Canonical domain set to biniou.fr
```

```text
$ scalingo -a biniou unset-canonical-domain
-----> Canonical domain disabled
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -19,7 +19,7 @@
 
 [[projects]]
   branch = "fix/cli/344/router_settings"
-  digest = "1:645b33e21dcebbf8ad42710ed3106429c3b04a595976178b4dbd89d71918d966"
+  digest = "1:65d0a0724e609db973ae32696314aab7890fbfe928f4e66710fb7a6dc67a23e0"
   name = "github.com/Scalingo/go-scalingo"
   packages = [
     ".",
@@ -28,7 +28,7 @@
     "io",
   ]
   pruneopts = "NUT"
-  revision = "349feb538faa51ae6c416de65fcf9d0cbc10ef50"
+  revision = "96dceb95a4c36550541d524c03b44614d0ff2b9d"
 
 [[projects]]
   branch = "master"
@@ -104,11 +104,11 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:8ba3f1b3c0cb6d601b3a8d3dfc3a899768614e77ff89a30c6eefdc34f75ff5ee"
+  digest = "1:e6baebf0bd20950e285254902a8f2241b4870528fb63fdc10460c67b1f31b1a3"
   name = "github.com/olekukonko/tablewriter"
   packages = ["."]
   pruneopts = "NUT"
-  revision = "b8a9be070da40449e501c3c4730a889e42d87a9e"
+  revision = "d4647c9c7a84d847478d890b816b7d8b62b0b279"
 
 [[projects]]
   digest = "1:5cf3f025cbee5951a4ee961de067c8a89fc95a5adabead774f82822efabab121"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -19,7 +19,7 @@
 
 [[projects]]
   branch = "fix/cli/344/router_settings"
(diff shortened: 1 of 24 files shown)


Gopkg.lock
```diff
@@ -3,107 +3,140 @@
...
 [[projects]]
   branch = "master"
+  digest = "1:e25669f49084df10ec9ff820eb625cfd90a136331bfb79112f88307f7829ad9f"
   name = "github.com/Scalingo/go-gitremote"
   packages = ["."]
+  pruneopts = "NUT"
   revision = "97b505dac0c3696d7395d6627494ec105a69280f"
 
 [[projects]]
+  branch = "fix/cli/344/router_settings"
```
johnsudaar: User the master branch

Gopkg.toml
```diff
@@ -25,8 +25,9 @@
   name = "github.com/Scalingo/envconfig"
 
 [[constraint]]
-  version = "~1.4"
   name = "github.com/Scalingo/go-scalingo"
+  branch = "fix/cli/344/router_settings"
```
johnsudaar: Same here

Gopkg.lock
EtienneM: Then merge https://github.com/Scalingo/go-scalingo/pull/71 :p

cmd/domains.go
```diff
@@ -145,4 +145,60 @@ var (
...
+	}
+
+	unsetCanonicalDomainCommand = cli.Command{
+		Name:     "unset-canonical-domain",
+		Category: "App Management",
+		Flags:    []cli.Flag{appFlag},
+		Usage:    "Unset a canonical domain.",
+		Description: `Unset the canonical domain of this app:
+
+    $ scalingo -a myapp unset-canonical-domain example.com
```
johnsudaar: Why not just `unset-canonical-domain` ? There is only one canonical domain per app.

EtienneM: :heavy_check_mark:

Soulou:
@johnsudaar is it good for you? (review says no)

johnsudaar:
Why? In this case shouldn't every actions prefixed by app:
`app:restart` `app:env-set` `app:scale`

Why this one and not the others ?

Soulou:
Hmm, maybe not app, but like `router:force-https` or `routing:force-https`, because stick-session lost in the numerous commands we have, it's not super clear, as we keep adding commands.

Soulou:
It's in the scope of this PR as we're adding some new commands.

EtienneM:
@Soulou Then do you agree if I rename all the argument so that we have:
`env-set` -> `env:set` etc for all commands?

Soulou:
There should be retrocompatibility for EVERYTHING. You don't break habits of people like this.

Soulou:
I completely agree that it's for another PR.

The question is what commands for this PR right now

* `scalingo settings-force-https --enable`
* `scalingo settings-canonical-domain --enable`

Just to have them gathered like in the dashboard.

johnsudaar:
Mmh we could move them in a new category called router, but i don't see the point in prefixing them. But i like the current implementation. I think that it suit the current CLI design.

Soulou:
Ok let's stop this horror story and let's merge this then.

PR #392, opened 2019-01-15
Add commands 'stacks' and 'stacks-set' to list and manipulate apps runtime stack

Soulou:


cmd/stacks.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/Scalingo/cli/stacks"
+	"github.com/urfave/cli"
+)
+
+var (
+	stacksListCommand = cli.Command{
+		Name:     "stacks",
+		Category: "Runtime Stacks",
+		Usage:    "List the available runtime stacks",
+		Description: ` List all the available rutime stacks for your apps:
```
EtienneM:
```suggestion
		Description: ` List all the available runtime stacks for your apps:
```

stacks/set.go
```diff
@@ -0,0 +1,39 @@
...
+
+func Set(app string, stack string) error {
+	c := config.ScalingoClient()
+	stacks, err := c.StacksList()
+	if err != nil {
+		return errgo.Notef(err, "fail to list available stacks")
+	}
+
+	var stackToSet scalingo.Stack
+	for _, apistack := range stacks {
```
EtienneM: Is it mandatory to do all this? `AppsSetStack` should already test that the stack exists before it sets it, shouldn't it?

Soulou: It's to get the ID, because we can't set a stack by its name.

Soulou:
Rebased

stacks/set.go
EtienneM: OK but don't we want to update the `AppsSetStack` method to accept both ID and name, like for some other API endpoint?

Soulou: Hmmm are we doing this? For which endpoint, can you tell me?

EtienneM: You're right, we don't do that. I supposed we do that for the apps endpoint in which we can use both the ID or the name but apparently no.

PR #395, opened 2019-02-22
New version of ssh package requrie HostKeyCallback, default value to ignore

### LINKED ISSUE #394: [ssh] Connection is broken with new SSH lib

Soulou:
Fixes #394

### DIFF
--- a/net/ssh/client.go
+++ b/net/ssh/client.go
@@ -72,8 +72,9 @@ func ConnectToSSHServer(keys []ssh.Signer) (*ssh.Client, ssh.Signer, error) {
 
 func ConnectToSSHServerWithKey(key ssh.Signer) (*ssh.Client, error) {
 	sshConfig := &ssh.ClientConfig{
-		User: "git",
-		Auth: []ssh.AuthMethod{ssh.PublicKeys(key)},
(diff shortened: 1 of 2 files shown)


net/ssh/client.go
```diff
@@ -72,8 +72,9 @@ func ConnectToSSHServer(keys []ssh.Signer) (*ssh.Client, ssh.Signer, error) {
 
 func ConnectToSSHServerWithKey(key ssh.Signer) (*ssh.Client, error) {
 	sshConfig := &ssh.ClientConfig{
-		User: "git",
-		Auth: []ssh.AuthMethod{ssh.PublicKeys(key)},
+		User:            "git",
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(key)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
EtienneM: From the documentaiton ( https://godoc.org/golang.org/x/crypto/ssh):

> InsecureIgnoreHostKey [...] should not be used for production code. 

Are you sure?...

EtienneM: Adding this flag is the exact same level of security as before updating the lib.

PR #396, opened 2019-02-28
Output only machine-parseable info on stdout

kephas:
Output human messages on stderr and ADDR:PORT on stdout for easier
machine parsing

### DIFF
--- a/db/tunnel.go
+++ b/db/tunnel.go
@@ -51,7 +51,7 @@ func Tunnel(opts TunnelOpts) error {
 	if err != nil {
 		return errgo.Notef(err, "invalid database 'URL': %s", dbUrlStr)
 	}
-	fmt.Printf("Building tunnel to %s\n", dbUrl.Host)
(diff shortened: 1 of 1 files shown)


db/tunnel.go
```diff
@@ -86,7 +86,8 @@ func Tunnel(opts TunnelOpts) error {
 	}
 
 	defer sock.Close()
-	fmt.Printf("You can access your database on '%v'\n", sock.Addr())
+	fmt.Println("You can access your database on:")
```
Soulou: Don't you want to display this on stderr also?

Soulou:
@kephas Thanks for the contribution, a small change to match the goal of the PR according to me

db/tunnel.go
kephas: Ugh, I used grep to test what goes in stdout but I did `grep ":"` and that line ends with one. <facepalm>

kephas: *facepalm*

PR #397, opened 2019-02-28
Add Backup download

johnsudaar:


johnsudaar:
All the code is there, I'm just waiting on https://github.com/Scalingo/go-scalingo/pull/76 to be merged.

cmd/backups.go
```diff
@@ -0,0 +1,72 @@
...
+			Usage: "Name of the current backup",
+		}, cli.StringFlag{
+			Name:  "output, o",
+			Usage: "Output file (- for stdout)",
+		}, cli.BoolFlag{
+			Name:  "silent, s",
+			Usage: "Do not show meter and loading messages",
+		}},
+		Description: `  Download a specific backup:
+		$ scalingo -a myapp --addon my_addon backup-download --backup my_backup
```
EtienneM: Please use long option instead of `-a`

cmd/backups.go
```diff
@@ -0,0 +1,72 @@
...
+
+		# See also 'backups' and 'addons'`,
+		Before: AuthenticateHook,
+		Action: func(c *cli.Context) {
+			currentApp := appdetect.CurrentApp(c)
+			addon := addonName(c)
+			backup := c.String("backup")
+			if backup == "" {
+				fmt.Println("Please specify an backup using the --backup flag")
```
EtienneM: `an backup` -> `a backup`

cmd/backups.go
```diff
@@ -0,0 +1,72 @@
...
+		Description: `  Download a specific backup:
+		$ scalingo -a myapp --addon my_addon backup-download --backup my_backup
+
+		# See also 'backups' and 'addons'`,
+		Before: AuthenticateHook,
+		Action: func(c *cli.Context) {
+			currentApp := appdetect.CurrentApp(c)
+			addon := addonName(c)
+			backup := c.String("backup")
+			if backup == "" {
```
EtienneM: Isn't it possible to tell the `cli` lib to make the option mandatory so that it automatically check that?

cmd/flags.go
```diff
@@ -0,0 +1,40 @@
...
+	var addonName string
+	if c.GlobalString("addon") != "<addon_id>" {
+		addonName = c.GlobalString("addon")
+	} else if c.String("addon") != "<addon_id>" {
+		addonName = c.String("addon")
+	} else if os.Getenv("SCALINGO_ADDON") != "" {
+		addonName = os.Getenv("SCALINGO_ADDON")
+	}
+	if addonName == "" {
+		fmt.Println("Unable to find the application name, please use --addon flag.")
```
EtienneM: `the application name` -> `the addon name`

config/auth/config.go
```diff
@@ -24,8 +24,12 @@ func NewConfigData() *ConfigData {
...
 type CredentialsData struct {
-	TokenGenerator *scalingo.APITokenGenerator `json:"tokens"`
-	User           *scalingo.User              `json:"user"`
+	Tokens *UserTokens    `json:"tokens"`
+	User   *scalingo.User `json:"user"`
 }
 
 type ConfigPerHostV2 map[string]*CredentialsData
+
+type UserTokens struct {
```
EtienneM: `UserTokens` -> `UserToken`

db/backup_download.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return errgo.Notef(err, "fail to start download")
+	}
+	defer resp.Body.Close()
+	spinner.Stop()
+	// Stop the spinner, start the progress bar
+	bar := pb.New(int(backup.Size)).SetUnits(pb.U_BYTES)
+	bar.Output = logWriter
+	bar.Start()
+	reader := bar.NewProxyReader(resp.Body) // Did I tell you this library is awesome ?
```
EtienneM: :heart:

db/backup_download.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if err != nil {
+		return errgo.Notef(err, "fail to download file")
+	}
+
+	return nil
+}
+
+// True if it's a valid path to a directory, false otherwise
```
EtienneM: A function comment in Golang starts with the name of the function. Here I would prefer `isDir returns true if...`

db/backups.go
```diff
@@ -0,0 +1,49 @@
...
+		return errgo.Notef(err, "fail to least backups")
+	}
+
+	t := tablewriter.NewWriter(os.Stdout)
+	t.SetHeader([]string{"ID", "Name", "CreatedAt", "Size", "Status"})
+
+	for _, backup := range backups {
+		t.Append([]string{
+			backup.ID,
+			backup.Name,
```
EtienneM: Will the download backup function accepts the name of the backup as input or just the ID? If it does not accept the name, is it really worth displaying it? Form my perspective there is a risk it creates confusion into our user mind.

cmd/backups.go
johnsudaar: I don't think so. I searched for it in the package doc and haven't found a way to add mendatory options.

db/backups.go
johnsudaar: Both

johnsudaar: Yeah you're right it might be confusing.

Gopkg.toml
```diff
@@ -26,7 +26,7 @@
 
 [[constraint]]
   name = "github.com/Scalingo/go-scalingo"
-  branch = "master"
+  branch = "fix/add_db_api"
```
EtienneM: I just wait for this to be merged in master then LGTM

cmd/backups.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+var (
+	BackupListCommand = cli.Command{
+		Name:     "backups",
+		Category: "Addons",
+		Usage:    "List backups for an addon",
+		Flags:    []cli.Flag{appFlag, addonFlag},
+		Description: `  List all available backups for an addon:
+		$ scalingo --app myapp --addon my_addon backups
```
Soulou: Isn't `my_addon`  going to be a UUID?

johnsudaar: It's already a UUID

Soulou: But I mean in the example?

PR #410, opened 2019-04-16
Display Let's Encrypt certificate status

### LINKED ISSUE #409: [domains] Let's Encrypt status error is not correctly handled
A domain on the web dashboard:

Screenshot from 2019-04-16 14-04-31

And the same on the CLI:

```
+----------------+--------------------------------+
|     DOMAIN     |              SSL               |
+----------------+--------------------------------+
| *.example.com   | Valid until 2019-04-12         |
|                | 19:40:07 +0200 CEST            |

```

EtienneM:
Fix #409

```
$ go build -o cli ./scalingo && ./cli -a my-app domains
+----------------+--------------------------------+
|     DOMAIN     |            TLS/SSL             |
+----------------+--------------------------------+
| example.com    | (LE In use) Valid until        |
|                | 2019-07-01 10:57:26 +0200 CEST |
| *.example.net  | (LE Error) Valid until         |
|                | 2019-04-12 19:40:07 +0200 CEST |
| *.example.eu   | (LE In use) Valid until        |
|                | 2019-06-12 09:00:11 +0200 CEST |
| example.net    |                              - |
| example.eu     |                              - |
+----------------+--------------------------------+

```

### DIFF
--- a/domains/list.go
+++ b/domains/list.go
@@ -38,16 +38,14 @@ func List(app string) error {
 		row := []string{domainName}
 		if !domain.SSL {
 			row = append(row, "-")
-		} else {
-			var tls string
-			if domain.LetsEncrypt {
-				letsencryptStatus, ok := letsencryptStatusString[string(domain.LetsEncryptStatus)]
-				if !ok {
-					letsencryptStatus = string(domain.LetsEncryptStatus)
-				}
-				tls = fmt.Sprintf("(LE %s)", letsencryptStatus)
+		} else if domain.LetsEncrypt {
+			letsencryptStatus, ok := letsencryptStatusString[string(domain.LetsEncryptStatus)]
+			if !ok {
+				letsencryptStatus = string(domain.LetsEncryptStatus)
 			}
-			row = append(row, fmt.Sprintf("%s Valid until %v", tls, domain.Validity))
+			row = append(row, fmt.Sprintf("Let's Encrypt: %s", letsencryptStatus))
+		} else {
+			row = append(row, fmt.Sprintf("Valid until %v", domain.Validity))
 		}
 		t.Append(row)
 	}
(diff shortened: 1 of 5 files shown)


domains/list.go
```diff
@@ -17,7 +26,7 @@ func List(app string) error {
 	}
 
 	t := tablewriter.NewWriter(os.Stdout)
-	t.SetHeader([]string{"Domain", "SSL"})
+	t.SetHeader([]string{"Domain", "TLS/SSL"})
```
johnsudaar: You could not resist? :stuck_out_tongue:

EtienneM: :laughing:

EtienneM:
I don't like it either, but I'm not sure a new column is better. How do you se this new column? Naming it `Let's Encrypt` and only adding it if there is at least a domain with a Let's Encrypt  certificate?

johnsudaar:
> I don't like it either, but I'm not sure a new column is better. How do you se this new column? Naming it `Let's Encrypt` and only adding it if there is at least a domain with a Let's Encrypt certificate?

Why not yep.

yannski:
Maybe the column should be named "TLS". Therefore Let's Encrypt status and Custom certificate status should be merged? BTW we have the same semantic "problem" on the web dashboard.

EtienneM:
```
$ go build -o cli ./scalingo && ./cli -a my-app domains
+----------------+-----------------------+
|     DOMAIN     |        TLS/SSL        |
+----------------+-----------------------+
| abcde.azert.me | Let's Encrypt: In use |
| *.azerty.net   | Let's Encrypt: Error  |
| *.azerty.eu    | Let's Encrypt: In use |
| azerty.net     |                     - |
| azerty.eu      |                     - |
+----------------+-----------------------+

```

johnsudaar:
And for custom certificates ?

EtienneM:
The code didn't change for custom certificate so it displays `Valid until  2019-07-01 10:57:26 +0200 CEST`

PR #425, opened 2019-06-22
Support multi-region

Soulou:


apps/run.go
```diff
@@ -182,7 +187,7 @@ func Run(opts RunOpts) error {
 		for {
 			select {
 			case s := <-signals:
-				run.HandleSignal(s, socket, ctx.attachURL)
+				run.HandleSignal(c, s, socket, ctx.attachURL)
```
EtienneM: Are you sure the Scalingo client can be shared by mutliple goroutines?

```suggestion
				run.HandleSignal(ctx.scalingoClient, s, socket, ctx.attachURL)
```

cmd/error.go
```diff
@@ -55,14 +55,24 @@ func (r *ReportError) Report() {
 }
 
 func errorQuit(err error) {
-	newReportError(err).Report()
+	currentUser, autherr := config.C.CurrentUser()
+	if autherr != nil {
+		debug.Println("Fail to get current user")
+		debug.Println(errgo.Details(err))
```
EtienneM:
```suggestion
		debug.Println(errgo.Details(err))
		return
```

Not sure whether you want to keep executing the function or not when you fail to get the user.

cmd/regions.go
```diff
@@ -1,27 +1,27 @@
...
 )
 
 var (
-	SignUpCommand = cli.Command{
-		Name:     "signup",
+	RegionsListCommand = cli.Command{
+		Name:     "regions",
 		Category: "Global",
-		Usage:    "Create your Scalingo account",
+		Usage:    "List available availability regions",
```
EtienneM:
```suggestion
		Usage:    "List available regions",
```

config/config.go
```diff
@@ -2,50 +2,73 @@ package config
...
+	env = map[string]string{
+		"SCALINGO_AUTH_URL":  "https://auth.scalingo.com",
+		"SCALINGO_API_URL":   "",
+		"SCALINGO_DB_URL":    "",
+		"SCALINGO_SSH_HOST":  "scalingo.com:22",
+		"SCALINGO_REGION":    "",
+		"API_VERSION":        "1",
+		"UNSECURE_SSL":       "false",
+		"ROLLBAR_TOKEN":      "",
+		"CONFIG_DIR":         ".config/scalingo",
```
EtienneM: Will it work properly on Windows or MacOS? Shouldn't we use each OS standard? Maybe through a lib such as https://github.com/shibukawa/configdir#configdir-for-golang?

Same question for the cache

regions/regions.go
```diff
@@ -0,0 +1,9 @@
+package regions
```
EtienneM: It looks like a useless file.

Soulou:
> Do we display somewhere the current region? I mean, if a user list his apps, how can he know on which zone he is? On the web dashboard, we have the zone selector for this. What about the CLI?

You're right, maybe `scalingo config` should display the current config of the CLI aka the region currently ?

Do you think we should display it when running a command also? Isn't that weird?

cmd/error.go
Soulou: Yes I'm sure it's just an HTTP Client finally, nothing has an internal state

regions/regions.go
Soulou: removed

config/config.go
Soulou: Actually I'm using exactly the same thing as we're doing for the authentication file which is also in .config/scalingo

There is some custom code for Windows/MacOS lower in that file

Soulou: That being said we're not exactly using the same PATH historically, but it would require a migration, I'm in favor of doing it in another issue, I'm opening it

Soulou: https://github.com/Scalingo/cli/issues/426

cmd/error.go
EtienneM: Your comment is related to another one of my comment. But you didn't addressed this one.

EtienneM:
> > Do we display somewhere the current region? I mean, if a user list his apps, how can he know on which zone he is? On the web dashboard, we have the zone selector for this. What about the CLI?
> 
> You're right, maybe `scalingo config` should display the current config of the CLI aka the region currently ?
> 
> Do you think we should display it when running a command also? Isn't that weird?

`scalingo config` to return the current region might be enough in the first place. Let's see if we have some support about some lost customers later.

cmd/error.go
Soulou: Oh my bad this was about the Client and goroutines right

Soulou:
Adding a column region in the app list?

EtienneM:
> Adding a column region in the app list?

When typing `scalingo apps`, we only display the apps in the current region, don't we? Adding a region column will just write N times the same information so I'm not really in favor of this. Does Heroku have multiple regions? Do they handle this use case?

Soulou:
@EtienneM Heroku output:

```
└> heroku apps
=== inscr@soulou.fr Apps
blooming-caverns-38122
radiant-sierra-63832
```

Soulou:
I'm going to implement `config` first and then we'll see

cmd/error.go
Soulou: Actually I want to keep on, at worse, the user will be nill for rollbar which is completely ok and taken into account

Soulou:
```
└> ./scalingo config
+-------------------+-----------+
| CONFIGURATION KEY |   VALUE   |
+-------------------+-----------+
| region            | agora-fr1 |
+-------------------+-----------+
``` 

```
└> ./scalingo regions      
+-----------+-------------+------------------------------------+
|   NAME    |   DISPLAY   |            API ENDPOINT            |
+-----------+-------------+------------------------------------+
| agora-fr1 | Paris (AC)  | https://api.agora-fr1.scalingo.com |
| osc-fr1   | Paris (OSC) | https://api.osc-fr1.scalingo.com   |
+-----------+-------------+------------------------------------+
```

PR #434, opened 2019-07-03
Remove dependency to gitremote

### LINKED ISSUE #432: Remove dependency to github.com/Scalingo/go-gitremote
`github.com/Scalingo/go-gitremote` is an old lib developed by Léo a long time ago and is not really well-tested. We started using `github.com/src-d/go-git` in other part of the CLI and should rather use this one.

EtienneM:
Fix #432


appdetect/git.go
```diff
@@ -51,42 +62,37 @@ func ScalingoRepoComplete(dir string) []string {
...
-			repos = append(repos, remote.Name)
+			repos = append(repos, remote.Config().Name)
 		}
 	}
 
 	return repos
 }
 
-func scalingoRemotes(directory string) (gitremote.Remotes, error) {
-	matchedRemotes := make(gitremote.Remotes, 0)
```
Soulou: Are you sure there is no regression here as the 'matchedRemotes' filter has been deleted?

Soulou:
"gopkg.in/src-d/go-git.v4/config" has not been added to dep/vendored

EtienneM:
> "gopkg.in/src-d/go-git.v4/config" has not been added to dep/vendored

It is already in the vendor folder

PR #443, opened 2019-07-11
Fix Git remote URL parsing to detect app name

### LINKED ISSUE #442: App detection based on Git config does not work anymore
https://app.intercom.io/a/apps/w4oogu7s/inbox/inbox/conversation/22798798439

The issue only occurs if the app is on the Outscale zone

EtienneM:
Fix #442

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### To be Released
 
+* Fix Git remote URL parsing to detect app name on Outscale [#443](https://github.com/Scalingo/cli/pull/443)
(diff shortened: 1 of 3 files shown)


Soulou:
It seems quite easy to test, can you add one?

EtienneM:
Not that easy as it calls `scalingoRemotes` which uses the Git lib to extract the Git repo information and we have no interface to mock it..

Soulou:
You can separate in two functions, and one is taking the name of the remote where you do the test?

PR #449, opened 2019-07-16
Only fill regions cache for regional commands

### LINKED ISSUE #447: Regions cache should only be filled for regional commands
And the `--password-only` flag is broken
And the help tries to login

This is because of this piece of code: https://github.com/Scalingo/cli/blob/master/cmd/commands.go#L32-L38

Before running any command, we fill the regions cache which requires to be logged in. And the call to `session.Login` do not use any flag on the command line.

EtienneM:
~Waiting for https://github.com/Scalingo/go-scalingo/pull/114 to be merged~ merged!

Fix #447

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,8 +10,7 @@
   revision = "8eb48cc6f27eafda8e3a9627edba494a1d229a01"
 
 [[projects]]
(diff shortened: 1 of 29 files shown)


Soulou:
LGTM

PR #455, opened 2019-07-22
Add periodic backups configuration commands

### LINKED ISSUE #454: It should be possible to configure db backup periodicity (enable/disable + hour)
https://github.com/Scalingo/go-scalingo/issues/118

EtienneM:
```
$ scalingo --app my-app --addon ad-d4903c5a-7b38-4fca-950a-54a34ae84ca4 backups-config --disable-scheduling
-----> Periodic backups are disabled

$ scalingo --app my-app --addon ad-d4903c5a-7b38-4fca-950a-54a34ae84ca4 backups-config --schedule-at 3
-----> Periodic backups will be done daily at 3:00 CEST
```

Fix #454

### DIFF
--- a/cmd/databases.go
+++ b/cmd/databases.go
@@ -1,0 +1,68 @@
+package cmd
+
+import (
+	"errors"
+
+	"github.com/Scalingo/cli/appdetect"
+	"github.com/Scalingo/cli/db"
+	"github.com/Scalingo/cli/io"
+	scalingo "github.com/Scalingo/go-scalingo"
+	"github.com/urfave/cli"
+)
+
+var (
+	databaseBackupsConfig = cli.Command{
+		Name:     "backups-config",
+		Category: "Addons",
+		Usage:    "Configure the periodic backups of a database",
+		Flags: []cli.Flag{appFlag, addonFlag, cli.IntFlag{
+			Name:  "scheduled-at",
+			Usage: "Hour of the day of the periodic backups (UTC)",
+		}, cli.BoolFlag{
+			Name:  "disable",
+			Usage: "Disable the periodic backups",
+		}, cli.BoolFlag{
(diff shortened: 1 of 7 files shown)


cmd/databases.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+var (
+	databaseBackupsConfig = cli.Command{
+		Name:     "backups-config",
+		Category: "Addons",
+		Usage:    "Configure the periodic backups of a database",
+		Flags: []cli.Flag{appFlag, addonFlag, cli.IntFlag{
+			Name:  "schedule-at",
+			Usage: "Enable daily backups and schedule them at the specified hour of the day (UTC)",
```
Soulou: Shouldn't we use the local timezone?

EtienneM: Yes it's probably better UX

Soulou:
And if no flag is used, shoudn't we display the current backups configuration?

yannski:
For the record here is how Heroku does it https://devcenter.heroku.com/articles/heroku-postgres-backups#scheduling-backups

cmd/databases.go
EtienneM: Here your are:

```
$ scalingo --app my-app --addon ad-d4903c5a-7b38-4fca-950a-54a34ae84ca4 backups-config --scheduled-at 3
-----> Periodic backups will be done daily at 3:00 CEST
```

Soulou:
@EtienneM what so you think, local time by default but ability to write "full tz" like heroku?

EtienneM:
> @EtienneM what so you think, local time by default but ability to write "full tz" like heroku?

Yep I will do that

And I like that Heroku name the option `unschedule`. I will also update this command name.

PR #463, opened 2019-08-02
Do not disconnect user on 401

### LINKED ISSUE #462: [Error Management] We should not disconnect the user if a 401 occured
https://github.com/Scalingo/cli/blob/master/cmd/error.go#L72

johnsudaar:
Fix #462


Soulou:
LGTM

PR #510, opened 2019-10-25
Fix creation of an integration link

### LINKED ISSUE #509: [Integration Link] Unable to create integration link for an app
Example :
```bash
scalingo -a gitlab-test integration-link-create https://gitlab.example.com/Scalingo/sample-go-martini
? Branch to auto-deploy (empty to disable): master
? Automatically deploy review apps: Yes
? Automatically destroy review apps when the pull/merge request is closed: Yes
? Hours before automatically destroying the review apps: 0
? Automatically destroy review apps after some time without deploy/commits: Yes
? Hours before automatically destroying the review apps: 0
 !     An error occured:
       fail to get the integration link for this app: fail to get this SCM repo link: The scm_repo_link was not found
```

It's for all SCM type not only GitLab.

brandon-welsch:
Fix #509

### DIFF
--- a/integrationlink/create.go
+++ b/integrationlink/create.go
@@ -28,10 +28,7 @@ func Create(app string, integrationType scalingo.SCMType, integrationURL string,
 		return errgo.Notef(err, "fail to get the integration")
 	}
 
-	repoLink, err := c.SCMRepoLinkShow(app)
-	if err != nil {
-		return errgo.Notef(err, "fail to get the integration link for this app")
-	}
+	repoLink, _ := c.SCMRepoLinkShow(app)
 	if repoLink != nil {
 		io.Statusf("Your app is already linked to %s/%s/%s", integration.URL, repoLink.Owner, repoLink.Repo)
 		if repoLink.Branch != "" {
--- a/integrationlink/create.go
+++ b/integrationlink/create.go
@@ -1,7 +1,6 @@
 package integrationlink
 
 import (
-	"errors"
 	"fmt"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


integrationlink/create.go
```diff
@@ -28,10 +28,7 @@ func Create(app string, integrationType scalingo.SCMType, integrationURL string,
 		return errgo.Notef(err, "fail to get the integration")
 	}
 
-	repoLink, err := c.SCMRepoLinkShow(app)
-	if err != nil {
```
EtienneM: I don't really like that you completely remove the error checking. Can't you just ignore the error in case it's a 404?

integrationlink/create.go
```diff
@@ -29,7 +30,7 @@ func Create(app string, integrationType scalingo.SCMType, integrationURL string,
 	}
 
 	repoLink, err := c.SCMRepoLinkShow(app)
-	if err != nil {
+	if scerr, ok := errors.ErrgoRoot(err).(*http.RequestFailedError); ok && scerr.Code != 404 {
```
EtienneM: What happens if the error type is not `http.RequestFailedError`?

EtienneM: Here, you are ignoring the error if it's a 404 or if it's anything else but a `http.RequestFailedError`. I think the intended behaviour is to ignore the error only in case of a 404, isn't it?

brandon-welsch: I return errors only if it's not 404, if the error is a 404 the code is executed normally. 

I don't understand, it's written in the code ^^

EtienneM: I mean that you MUST return if the error is not a `http.RequestFailedError`. And it's not the case.

brandon-welsch: Hmmm ok

PR #517, opened 2020-01-09
Fix: do not fail on short connections

### LINKED ISSUE #516: [db-tunnel] On ES, the connection is broken after each call
The db-tunnel has been made for long standing connections. However, ES uses HTTP to communicate so connections are short lived. This lead to bad user experience since after the first request is made, the db-tunnel consider the connection as broken (since both the client and the server has closed the connection) and tries to reconnect (with a 10s cooldown).
Even worse: the reconnection fails

```
scalingo --region osc-fr1 -a kibana db-tunnel SCALINGO_ELASTICSEARCH_URL                                                                      
Building tunnel to REDACTED.elasticsearch.dbs.scalingo.com:123456                                                                                                        
You can access your database on:                                                     
127.0.0.1:10000                           
Connect to [REDACTED].elasticsearch.dbs.scalingo.com:123456 [1]       
// HERE THE HTTP CONNECTION IS MADE                                                                                                     
End of connection [1]                     
Connection broken, reconnecting...                                                   
Fail to reconnect, waiting 10 seconds...                                             
Fail to reconnect, waiting 10 seconds...                                             
Fail to reconnect, waiting 10 seconds...   
```

johnsudaar:
Fix #516

* Do not reuse ssh.Signer, the connection to ssh agent might be broken
* Do not retry the connection if the server closed the connection
* Add --bind to db-tunnel to bind a custom command. This is helpful for docker setups where listening on localhost wont make it available on the docker0 interface

### DIFF
--- a/db/tunnel.go
+++ b/db/tunnel.go
@@ -1,7 +1,6 @@
 package db
 
 import (
-	"errors"
 	"fmt"
 	stdio "io"
 	"net"
@@ -22,7 +21,6 @@ import (
 )
 
 var (
-	errTimeout      = errors.New("timeout")
 	connIDGenerator = make(chan int)
 	defaultPort     = 10000
 	defaultBind     = "127.0.0.1"
@@ -215,7 +213,7 @@ func handleConnToTunnel(sshClient *ssh.Client, dbUrl *url.URL, sock net.Conn, er
 			return nil
 		}
 
-		return errTimeout
+		return err
 	}
 	return nil
 }
--- a/db/tunnel.go
+++ b/db/tunnel.go
@@ -13,10 +13,10 @@ import (
 	"time"
 
 	"github.com/Scalingo/cli/config"
-	"github.com/Scalingo/go-scalingo/debug"
 	"github.com/Scalingo/cli/io"
 	netssh "github.com/Scalingo/cli/net/ssh"
 	"github.com/Scalingo/go-scalingo"
+	"github.com/Scalingo/go-scalingo/debug"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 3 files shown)


db/tunnel.go
```diff
@@ -198,6 +207,14 @@ func handleConnToTunnel(sshClient *ssh.Client, dbUrl *url.URL, sock net.Conn, er
 	fmt.Printf("End of connection [%d]\n", connID)
 	// Connection timeout
 	if err != nil && strings.Contains(err.Error(), "use of closed network") {
+
+		// If the connection has been closed by the CLIENT, we must stop here and return a nil error
+		// If the connection has been closed by the SERVER, we must return a errTimeout and retry the connection
```
Soulou: Isn't it legitimate that the server closes the connection sometimes?

db/tunnel.go
```diff
@@ -198,6 +207,14 @@ func handleConnToTunnel(sshClient *ssh.Client, dbUrl *url.URL, sock net.Conn, er
...
+		clientClosedConnectionTest := fmt.Sprintf("%s->%s: use of closed network connection", sock.LocalAddr(), sock.RemoteAddr()) // Golang error checking 101 <3
+		if strings.Contains(err.Error(), clientClosedConnectionTest) {
+			return nil
+		}
+
 		return errTimeout
```
Soulou: Why not returning `err`?

db/tunnel.go
johnsudaar: Not on this connection, if the server closed the conneciton, it will break the conneciton at line 191.  Here it's the ssh conneciton that broke.

Soulou:
LGTM

PR #519, opened 2020-02-04
Error message if unknown app suggest to try on a different region

EtienneM:
```
╰─$ go build -o scalingo-cli ./scalingo && ./scalingo-cli --app unknown-app run bash
 !     The application was not found on the region agora-fr1. You can try on a different region with 'scalingo --region osc-fr1 ...'.
```

Related to https://github.com/Scalingo/one-stop-shop/issues/34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### To be Released
 
 * Add `--bind` arg to `db-tunnel` command that let us bind a custom host (and not only 127.0.0.1) [#517](https://github.com/Scalingo/cli/pull/517)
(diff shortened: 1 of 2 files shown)


EtienneM:
The two lines version looks, indeed, much better:

```
╰─$ go build -o scalingo-cli ./scalingo && ./scalingo-cli --app unknown-app run bash
 !     The application was not found on the region agora-fr1.
 !     You can try on a different region with 'scalingo --region osc-fr1 ...'.
```

mackwic:
So eager to see it live ! :D

PR #521, opened 2020-02-11
Fix error management in integration-link-create

### LINKED ISSUE #520: [ScmRepoLink] Cannot link an app to a repo link
When checking if the app is already linked, we use the following condition:

```
if scerr, ok := errors.ErrgoRoot(err).(*http.RequestFailedError); ok || (ok && scerr.Code != 404) {
	return errgo.Notef(err, "fail to get the integration link for this app")
}
```

Howerver if there is a 404 error (which is likely since the app should not be linked), err will be something like:

```
&http.RequestFailedError{
  Code: 404
}
```

So ok will be true. 

Since the condition for failure is:
```
ok || (ok && scerr.Code != 404)
```

This will be `true || false` => true and we will fail.

johnsudaar:
Fix #520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### To be Released
 
+* Bugfix: integration-link-create: command will always fail if there's already a link [#520](https://github.com/Scalingo/cli/issues/520)
+
 ### 1.16.5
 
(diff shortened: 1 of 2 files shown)


Soulou:
LGTM

PR #524, opened 2020-02-13
Improve error message when querying unknown app

EtienneM:
```
╰─$ go build -o scalingo-cli ./scalingo && ./scalingo-cli --app unknown-app run bash
 !     The application was not found on the region agora-fr1.
 !     You can try on a different region with 'scalingo --region osc-fr1 ...'.
 !     
 !     List of available regions from your account is accessible with 'scalingo regions'.
```

Related to https://github.com/Scalingo/one-stop-shop/issues/34

### DIFF
--- a/cmd/error.go
+++ b/cmd/error.go
@@ -84,6 +84,8 @@ func errorQuit(err error) {
 				region := strings.Split(apiURL.Host, ".")[1]
 				io.Errorf("The application was not found on the region %s.\n", region)
 				io.Error("You can try on a different region with 'scalingo --region osc-fr1 ...'.")
+				io.Error("")
+				io.Error("List of available regions is accessible with 'scalingo regions'.")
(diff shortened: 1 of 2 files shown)


cmd/error.go
```diff
@@ -84,6 +84,8 @@ func errorQuit(err error) {
 				region := strings.Split(apiURL.Host, ".")[1]
 				io.Errorf("The application was not found on the region %s.\n", region)
 				io.Error("You can try on a different region with 'scalingo --region osc-fr1 ...'.")
+				io.Error("")
+				io.Error("List of available regions from your account is accessible with 'scalingo regions'.")
```
johnsudaar:
```suggestion
				io.Error("List of available regions for your account is accessible with 'scalingo regions'.")
```
It makes more sense for me. Wdyt ?

PR #527, opened 2020-03-11
Fix/cli/525/git setup remote already exists

### LINKED ISSUE #525: [Git-Setup] This command should not fail if the remote already exist
If the remote already exist, it should update it instead of just complaining.

cyrille-sg:
Offer a option --force to the command git-setup to allow the user to override the current remote handle with a new remote url

Fix #525 

User
- New flag "--force" on "git-seup"
- Suggestion message to use force when already remote handle already exist
Tech
- Propose change in errors management (using utils.Wrap instead of in place errgo.Notef)

### DIFF
--- a/cmd/git.go
+++ b/cmd/git.go
@@ -16,6 +16,9 @@ var (
 			cli.StringFlag{
 				Name: "remote, r", Value: "scalingo",
 				Usage: "Specify the remote name"},
+			cli.BoolFlag{
+				Name:  "force, f",
+				Usage: "Replace remote url even if remote already exist"},
 		},
 		Description: `Add a Git remote to the current folder.
 
@@ -29,7 +32,8 @@ var (
 			}
 
 			err := git.Setup(appdetect.CurrentApp(c), git.SetupParams{
-				RemoteName: c.String("r"),
+				RemoteName:     c.String("r"),
+				ForcePutRemote: c.Bool("f"),
 			})
 			if err != nil {
 				errorQuit(err)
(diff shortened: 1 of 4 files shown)


cmd/git.go
```diff
@@ -29,7 +32,8 @@ var (
 			}
 
 			err := git.Setup(appdetect.CurrentApp(c), git.SetupParams{
-				RemoteName: c.String("r"),
+				RemoteName:     c.String("r"),
```
Soulou: Please use the long flag name for a better readability

git/setup.go
```diff
@@ -24,14 +26,66 @@ func Setup(appName string, params SetupParams) error {
...
+func createRemoteInRepository(repository *git.Repository, remoteName string, url string) error {
+	_, err := repository.CreateRemote(&gitconfig.RemoteConfig{
+		Name: remoteName,
+		URLs: []string{url},
+	})
+	if err != nil {
+		errWrapped := utils.WrapError(err, "fail to create the Git remote")
+		if err == git.ErrRemoteExists {
+			errWrapped = utils.WrapError(err, "fail to create the Git remote"+
+				" (maybe you want use the force option to override the remote url)")
```
Soulou:
```
Fail to configure git repository, "%v" remote already exists. (Use --force option to override)
```

cyrille-sg:
Screenshot from 2020-03-11 11-39-07

Soulou:
As spoken orally, the error displayed to the user is not right (repetition), to make it cleaner, it requires some changes in way we're handling error and probably to modify https://github.com/Scalingo/go-scalingo/errors

Soulou:
After disucssion, we accept this version, better error management will be handled later

PR #540, opened 2020-03-27
Display region name in migration example commands

### LINKED ISSUE #539: [Migration] Region should be displayed in command examples

Soulou:
Fixes #539


Soulou:
Done

PR #549, opened 2020-04-20
[Migration] Explain what's the current state is doing

### LINKED ISSUE #547: [migration-follow] Make it clearer the it's normal behaviour to be pending
The `migration-follow` command stay pending, waiting for new events to happen in the given migration. It's unclear that it's the normal behaviour. I suggest two improvements:
- The help message just states "Follow a running migration". Clarify that it will display the update to the migration as they arrive and it's pending waiting for these events.
- Display a spinner to highlight the _pending_ nature of this command with a message like "waiting for update to this migration" or something like that.

https://app.intercom.com/a/apps/w4oogu7s/inbox/inbox/all/conversations/26478545729

johnsudaar:
Fix #547

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -26,7 +26,7 @@
   revision = "8eb48cc6f27eafda8e3a9627edba494a1d229a01"
 
 [[projects]]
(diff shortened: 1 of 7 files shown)


johnsudaar:
Some examples

PR #550, opened 2020-04-20
[Migration] Retry if refresh errored

### LINKED ISSUE #546: [Migrations] fail to refresh migration: it should be invisible/retried

johnsudaar:
Fix #546

### DIFF
--- a/region_migrations/refresher.go
+++ b/region_migrations/refresher.go
@@ -27,6 +27,7 @@ type Refresher struct {
 
 	lock                 *sync.Mutex
 	migration            *scalingo.RegionMigration
+	errCount             int
 	stop                 bool
 	screenRefreshTime    time.Duration
 	migrationRefreshTime time.Duration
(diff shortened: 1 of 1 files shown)


johnsudaar:
If it fails during the load.


Then it works


Then it re-fail

region_migrations/refresher.go
```diff
@@ -93,16 +96,28 @@ func (r *Refresher) migrationRefresher() error {
 	client := r.client
 	r.lock.Unlock()
 
+	errCount := 0
+
+	var migration *scalingo.RegionMigration
+
 	for {
```
Soulou, line 103: Could you use `go-utils/retry` ?

johnsudaar: I thought about it and it wont be easy here. There's no mechanisme to know where we are in the retry. And the code still need to run. This will need a major redesign of this part and to add very different functionnalities to this package. Not sure this is worth the time.

Soulou: It would be at least interesting to list missing features in the package.

Like a callback with the iteration number and some other information or something like that, don't you think?

Soulou: (I'd be glad to have a quite polyvalent retry package)

johnsudaar: Ok i'll see

PR #557, opened 2020-04-21
Use new retrier to manage refresh tries

### LINKED ISSUE #545: [migration] we should mention the update of the git address

johnsudaar:


johnsudaar:
Same UI as before (a bit more reactive)

PR #567, opened 2020-05-25
Log drains: delete command

### LINKED ISSUE #554: Remove a drain
Remove a drain by application

```bash
scalingo --app my-app logs-drains-remove <id>
```

curzolapierre:
Add remove command

Example:
```bash
$ scalingo log-drains-remove --app sample-go-martini tcp+tls://logs.papertrailapp.com:12345
-----> The log drain: tcp+tls://logs.papertrailapp.com:12345 has been deleted
```

Fix #554

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -27,7 +27,7 @@
 
 [[projects]]
   branch = "feature/cli/554/add_log_drains_del_cmd"
-  digest = "1:de17197855c36bda72c0fee89959ca4f745e893a5185f3f9f26dab316975f13b"
+  digest = "1:a1635c85bc0f760bbc43b180f4dd51e04d8a1774752268cb555446435c46b402"
   name = "github.com/Scalingo/go-scalingo"
   packages = [
     ".",
@@ -38,7 +38,7 @@
     "io",
   ]
   pruneopts = "NUT"
-  revision = "b83c1952fb9b04dc04e0e4281d0e94cc2b366240"
+  revision = "ea8f03c8a69643e438a23d25afc24eb0b3686498"
 
 [[projects]]
   digest = "1:ab7f2d565135ccc3b35eb2427819fbca0501d582eba5091e684275022d3ff3ba"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -27,7 +27,7 @@
 
 [[projects]]
   branch = "feature/cli/554/add_log_drains_del_cmd"
-  digest = "1:a1635c85bc0f760bbc43b180f4dd51e04d8a1774752268cb555446435c46b402"
+  digest = "1:07201e6d503d992dfe159527dac247af0a27b3ec608e958d10ba77b98055f708"
   name = "github.com/Scalingo/go-scalingo"
   packages = [
     ".",
@@ -38,7 +38,7 @@
     "io",
   ]
   pruneopts = "NUT"
-  revision = "ea8f03c8a69643e438a23d25afc24eb0b3686498"
+  revision = "0f54664864f4de65ff77cb47eb1eb79eaf16a63b"
 
 [[projects]]
   digest = "1:ab7f2d565135ccc3b35eb2427819fbca0501d582eba5091e684275022d3ff3ba"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -27,7 +27,7 @@
 
 [[projects]]
   branch = "feature/cli/554/add_log_drains_del_cmd"
-  digest = "1:40ae488a4e927a7c247abd84487118d38f19bbd400b65747d2818fa03424afb7"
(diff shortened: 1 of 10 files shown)


Soulou:
Why does the user need to use an URL? Didn't we agree, it should be "easier", using flags etc.?

curzolapierre:
@Soulou We planned to use uuid for the MVP. But as we are only on the first step of the MVP here, without uuid implemented yet, we (with @EtienneM) decided to use urls returned by the list command.
I agree on the point it is less consistent to use url, but flags are usefull to control url generation and are less relevant in case of deletion.
What do you think ?

Soulou:
So if I understand this well, the problem is that we don't have a unique identifier for each log drain, so we can't ask users to use them for the deletion, right?

curzolapierre:
I hadn't thought about that, the logs service will allow to add multiple same url by application ?
The current implementation of addition on the API will return an error if the user try to add an existing url.

cmd/autocomplete/log_drains_remove.go
```diff
@@ -0,0 +1,30 @@
...
+		return nil
+	}
+
+	client, err := config.ScalingoClient()
+	if err != nil {
+		return errgo.Notef(err, "fail to get Scalingo client")
+	}
+	drains, err := client.LogDrainsList(appName)
+	if err == nil {
+
```
EtienneM, line 23:
```suggestion
```

cmd/autocomplete/log_drains_remove.go
```diff
@@ -0,0 +1,30 @@
...
+	if appName == "" {
+		return nil
+	}
+
+	client, err := config.ScalingoClient()
+	if err != nil {
+		return errgo.Notef(err, "fail to get Scalingo client")
+	}
+	drains, err := client.LogDrainsList(appName)
+	if err == nil {
```
EtienneM, line 22: We usually compare `err` with `!=nil`. It's odd to do the other way around. Can you modify this piece of code to compare `if err != nil` first? And add a comment to explain why you ignore the error please

cmd/log_drains.go
```diff
@@ -80,4 +80,34 @@ var (
...
+	logDrainsRemoveCommand = cli.Command{
+		Name:     "log-drains-remove",
+		Category: "Log drains",
+		Flags:    []cli.Flag{appFlag},
+		Usage:    "Remove a log drain from an application",
+		Description: `Remove a log drain from an application:
+
+	$ scalingo --app my-app log-drains-remove syslog://custom.logstash.com:12345
+
+	# See also commands 'log-drains-add', 'log-drains' and 'log-drains'`,
```
EtienneM, line 93:
```suggestion
	# See also commands 'log-drains-add', 'log-drains'`,
```

log_drains/remove.go
```diff
@@ -0,0 +1,22 @@
...
+
+func Remove(app string, URL string) error {
+	c, err := config.ScalingoClient()
+	if err != nil {
+		return errgo.Notef(err, "fail to get Scalingo client")
+	}
+
+	err = c.LogDrainRemove(app, URL)
+	if err != nil {
+		return errgo.Mask(err)
```
EtienneM, line 17:
```suggestion
		return errgo.Notef(err, "fail to remove the log drain from the application")
```

log_drains/remove.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/Scalingo/cli/config"
+	"github.com/Scalingo/cli/io"
+	"gopkg.in/errgo.v1"
+)
+
+func Remove(app string, URL string) error {
+	c, err := config.ScalingoClient()
+	if err != nil {
+		return errgo.Notef(err, "fail to get Scalingo client")
```
EtienneM, line 12:
```suggestion
		return errgo.Notef(err, "fail to get Scalingo client to remove a log drain from the application")
```

cmd/autocomplete/log_drains_remove.go
curzolapierre: Silly copy past from another remove autocomplete my bad. I'll fix the other too

Gopkg.toml
```diff
@@ -26,7 +26,7 @@
 
 [[constraint]]
   name = "github.com/Scalingo/go-scalingo"
-  version = "^4"
+  branch = "feature/cli/554/add_log_drains_del_cmd"
```
EtienneM, line 29:
```suggestion
  version = "^4"
```

Gopkg.lock
```diff
@@ -37,8 +38,7 @@
     "io",
   ]
   pruneopts = "NUT"
-  revision = "9f30cefb0ea118d42f84c30964bbb3a842f004be"
-  version = "v4.5.2"
```
EtienneM, line 41: I think you didn't run `dep ensure -update github.com/Scalingo/go-scalingo` as this line should be updated to `v4.5.3`

PR #589, opened 2020-08-27
[dev] Add a PR template

EtienneM:
I think we often forget to add the chngelog entry. I hope that this change will make it less forgettable.


Soulou (approved):
LGTM

PR #592, opened 2020-08-31
Fetch the ssh url from our region config

### LINKED ISSUE #591: The CLI is still using ssh.scalingo.com:22 for login
This endpoint is deprecated and we should use the region metadata to get those kind of URLs.

johnsudaar:
Fix #591


Soulou (approved):
LGTM

PR #599, opened 2020-09-02
Make an unauthenticated request to fetch the region cache on login

### LINKED ISSUE #598: Fail to login via SSH if the region cache is not present
If a user tries to login via SSH when the region cache is not present, we will try to make an authenticated request to get the region list wich fails since the user is not authenticated.

johnsudaar:
Fix #598

### DIFF
--- a/config/region.go
+++ b/config/region.go
@@ -34,7 +34,8 @@ func (c RegionsCache) Default() (scalingo.Region, error) {
 // GetRegionOpts allows the caller to use a custom API token instead of
 // the default one of the authentication used
 type GetRegionOpts struct {
-	Token string
+	Token    string
+	SkipAuth bool
 }
 
 func EnsureRegionsCache(c Config, opts GetRegionOpts) (RegionsCache, error) {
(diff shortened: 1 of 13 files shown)


johnsudaar:
Waiting for https://github.com/Scalingo/go-scalingo/pull/180

config/region.go
```diff
@@ -50,27 +51,41 @@ func EnsureRegionsCache(c Config, opts GetRegionOpts) (RegionsCache, error) {
...
 	regions, err := client.RegionsList()
 	if err != nil {
 		return RegionsCache{}, errgo.Notef(err, "fail to list available regions")
 	}
 
 	regionsCache.Regions = regions
 	regionsCache.ExpireAt = time.Now().Add(10 * time.Minute)
+
+	if opts.SkipAuth {
+		return regionsCache, nil
```
EtienneM, line 86: Why don't you write the regions cache in case you skip the auth? I have my idea on this matter but it might be interesting to explain it in a comment here.

EtienneM (approved):

PR #632, opened 2021-03-24
feat(ps): add ps command to list all application containers

EtienneM:
Related to [OPSB-253]


apps/ps.go
```diff
@@ -0,0 +1,31 @@
...
+	containers, err := c.AppsPs(app)
+	if err != nil {
+		return errgo.Notef(err, "fail to list the application containers")
+	}
+
+	t := tablewriter.NewWriter(os.Stdout)
+	t.SetHeader([]string{"Name", "Status", "Command", "Size", "Created At"})
+
+	for _, container := range containers {
+		t.Append([]string{container.Label, container.State, container.Command, container.Size, container.CreatedAt.Format(utils.TimeFormat)})
```
Soulou, line 27: Could you also to the update to stop using `container.Size` but use the `container_size` model ?

Soulou (approved):

PR #644, opened 2021-04-01
fix(ps): command needs an extra API query for web containers

EtienneM:
And probably for every non-one-off containers, but I didn't checked

The command was empty before this PR.

```
╰─$ go build -o scalingo-cli ./scalingo && ./scalingo-cli -a biniou ps
+-------+---------+-------------------+------+---------------------+
| NAME  | STATUS  |      COMMAND      | SIZE |     CREATED AT      |
+-------+---------+-------------------+------+---------------------+
| web-1 | running | sample-go-martini | M    | 2021/03/26 16:49:05 |
+-------+---------+-------------------+------+---------------------+
```

Related to [OPSB-253]

[OPSB-253]: https://scalingo.atlassian.net/browse/OPSB-253

### DIFF
--- a/apps/ps.go
+++ b/apps/ps.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/Scalingo/cli/config"
 	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v4"
 	"github.com/olekukonko/tablewriter"
 	"gopkg.in/errgo.v1"
 )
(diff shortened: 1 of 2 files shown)


Soulou (changes requested):
Why isn't the `Command` returned by the `Container` model in the API, I'm not a fan in this mapping while there is a `command` field in the container.

https://github.com/Scalingo/go-scalingo/blob/master/container.go#L22

EtienneM:
As stated in the description of the PR, this is empty for web containers, and probably for everything that is not a one-off. This is how things are stored on the API:

```
[6] pry(main)> app.containers.find_by(type: "web").command
=> ""
```

I wanted to circumvent this. Do you think it should be done in the API?

Soulou:
> As stated in the description of the PR, this is empty for web containers, and probably for everything that is not a one-off. This is how things are stored on the API:
> 
> ```
> [6] pry(main)> app.containers.find_by(type: "web").command
> => ""
> ```
> 
> I wanted to circumvent this. Do you think it should be done in the API?

I think our code should be fixed and the `command` should be denormalized always when the `Container` is created → In our scheduler

Soulou (commented):
Commented above ^

EtienneM:
OK I'm opening an issue on the scheduler and closing this PR.

Issue #609, opened 2021-01-06
Add support for the new Apple M1 architecture


EtienneM:
Installing the CLI on the new Macbook fails:

```
curl -O https://cli-dl.scalingo.com/install && bash install

% Total % Received % Xferd Average Speed Time Time Time Current
Dload Upload Total Spent Left Speed
100 3689 0 3689 0 0 11528 0 --:--:-- --:--:-- --:--:-- 11528
-----> Downloading Scalingo client... DONE
-----> Extracting... Archive: /tmp/scalingo_cli_YKm/scalingo_1.19.3_darwin_arm64.zip
End-of-central-directory signature not found. Either this file is not
a zipfile, or it constitutes one disk of a multi-part archive. In the
latter case the central directory and zipfile comment will be found on
the last disk(s) of this archive.
unzip: cannot find zipfile directory in one of /tmp/scalingo_cli_YKm/scalingo_1.19.3_darwin_arm64.zip or
/tmp/scalingo_cli_YKm/scalingo_1.19.3_darwin_arm64.zip.zip, and cannot find /tmp/scalingo_cli_YKm/scalingo_1.19.3_darwin_arm64.zip.ZIP, period.
DONE
-----> Install scalingo client to /usr/local/bin
sudo required...
mv: rename /tmp/scalingo_cli_YKm/scalingo_1.19.3_darwin_arm64/scalingo to /usr/local/bin/scalingo: No such file or directory
! Fail to install scalingo client (return 1)
```

It seems that the version targetted to amd64 works.

https://app.intercom.com/a/apps/w4oogu7s/inbox/inbox/all/conversations/12375700014217

EtienneM:
We use [goxc](https://github.com/laher/goxc) for the cross-compilation. We may just have to customize the `.goxc.json` file to add this architecture.

EtienneM:
This official [blog article](https://blog.golang.org/ports) states:

> Earlier this week, we released the first Go 1.16 beta, which includes native support for Macs using the M1 chip.

Maybe we should just wait for Go 1.16 release :)

czj:
I think waiting for a new Go release would be great @EtienneM but in the meantime via https://github.com/Scalingo/cli/issues/611 you could have the amd64 / x86 binary work on arm64 Macs

PR #648, opened 2021-04-20
dists(build) Build releases for arm64 to also build for Apple M1 architecture

Soulou:


EtienneM (commented):
DOes it fix #609?

Soulou:
It doesn't fix completely since the install script may have to be changed, BUT, the binary produced work correctly on Apple M1, a customer has tested with success

EtienneM (approved):

PR #652, opened 2021-05-05
fix: remove TODO

EtienneM:
I think these TODOs were never an actual issue. Let's remove them and open issue later if it's needed

- [ ] Add a changelog entry in the section "To Be Released" of CHANGELOG.md: no need for a changelog entry here I think


curzolapierre (approved):
LGTM

PR #659, opened 2021-05-27
fix(logs) Buffer logs to prevent requests from timing out

### LINKED ISSUE #658: [Logs] Missing logs
Our logs colorization pipeline is slow. This can make request silently timeout and last logs will be missing from the output result.
We should buffer the logs response body as fast as possible and do not block the reads from the colorization.

johnsudaar:
Fix #658

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -8,6 +8,7 @@ import (
 	"os/signal"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 	"unicode"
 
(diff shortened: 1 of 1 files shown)


johnsudaar:
> Wow, nicely caught!!! I assume this is related to the customer report from yesterday?
Thanks :) It took us some time to undestand what happened with @Soulou but we finally got it !

> I think we should do a release of the CLI so that this fix is out as soon as possible. Do you handle it or should I?
Yes if we could release a new version of the CLI soon it would much appreciated. If you have time i would love if you could do it.

EtienneM:
OK I'll do it asap, tomorrow max

johnsudaar:
Thanks !

PR #663, opened 2021-06-02
fix(backup-download) Now download the last successful backup

### LINKED ISSUE #657: [backups-download] Should download the latest successfull backup
It currently downloads the latest backup without argument. But if the latest backup is in error state, no need to try to download it, our API will return an error.

TheoEb:
Add loops to range over the last backups until we find a successful one to download.
Still return an error if there is none.

Fix #657

### DIFF
--- a/db/backup_download.go
+++ b/db/backup_download.go
@@ -43,8 +43,6 @@ func DownloadBackup(app, addon, backupID string, opts DownloadBackupOpts) error 
 	if err != nil {
 		return errgo.Notef(err, "fail to get Scalingo client to download a backup")
 	}
-
-	backupIDs := []string{}
 	if backupID == "" {
 		backups, err := client.BackupList(app, addon)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


db/backup_download.go
```diff
@@ -50,8 +53,12 @@ func DownloadBackup(app, addon, backupID string, opts DownloadBackupOpts) error
 		if len(backups) == 0 {
 			return errgo.New("This addon has no backup")
 		}
-		backupID = backups[0].ID
+		for _, backup := range backups {
+			backupIDs = append(backupIDs, backup.ID)
+		}
```
EtienneM, line 58: I wouldn't do it that way. I would call here a method, for instance:

```go
backup, err := getLastSuccessfulBackup(ctx, backups)
```

And that's it for the modification in this method. 

Here, you modified the behaviour of the method when a customer calls `backup-download` with a backup ID in argument. But we don't want to modify this behaviour.

EtienneM (changes requested):
Your commits are unverified because "The email in this signature doesn’t match the committer email." 
Do you have the commiter email in your GitHub profile?

db/backup_download.go
```diff
@@ -131,3 +135,12 @@ func isDir(path string) bool {
...
 	return s.IsDir()
 }
+
+func getLastSuccessfulBackup(backups []scalingo.Backup) (string, error) {
+	for _, backup := range backups {
+		if backup.Status == scalingo.BackupStatusDone {
+			return backup.ID, nil
+		}
+	}
+	return "", errgo.New("Can't find any successful backup")
```
EtienneM, line 145: Error message starts with a lower case in Go (https://www.notion.so/scalingooriginal/Golang-Best-Practice-de47c3b8d6c24c2e9b70c7b9c2f36660#76e817f954bb46ed85adb08bfd850f83)
```suggestion
	return "", errgo.New("can't find any successful backup")
```

EtienneM (changes requested):

EtienneM (approved):

PR #665, opened 2021-06-14
Fix(download backup) Print error on stderr if the backup status is no…

TheoEb:
…t 'done'

Linked to https://github.com/Scalingo/database-api-service/pull/243
Fix https://github.com/Scalingo/database-api-service/issues/229

### DIFF
--- a/db/backup_download.go
+++ b/db/backup_download.go
@@ -104,17 +104,17 @@ func DownloadBackup(app, addon, backupID string, opts DownloadBackupOpts) error 
 	}
 	defer resp.Body.Close()
 
-	// Return an error if the api return 400
+	// Stop the spinner
+	spinner.Stop()
+
 	if resp.StatusCode != http.StatusOK {
-		spinner.Stop()
 		return httpclient.NewRequestFailedError(resp, &httpclient.APIRequest{
 			URL:         downloadURL,
 			Method:      "GET",
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


db/backup_download.go
```diff
@@ -102,6 +103,16 @@ func DownloadBackup(app, addon, backupID string, opts DownloadBackupOpts) error
 		return errgo.Notef(err, "fail to start download")
 	}
 	defer resp.Body.Close()
+
+	// Return an error if the api return 400
```
EtienneM, line 107: The comment does not add anything that the code already says, moreover it is wrong

```suggestion
```

EtienneM (approved):

db/backup_download.go
```diff
@@ -102,6 +103,15 @@ func DownloadBackup(app, addon, backupID string, opts DownloadBackupOpts) error
 		return errgo.Notef(err, "fail to start download")
 	}
 	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		spinner.Stop()
```
EtienneM, line 108: Sorry I didn't noticed it at the first review, but can you move this line before the `if`, and remove the call to `spinner.STop` on line 115?

EtienneM (changes requested):

EtienneM (approved):

PR #674, opened 2021-08-23
fix(run): allow an equals sign (=) in env var values

### LINKED ISSUE #673: `scalingo run`: can't use `=` in environment variable values
I tried to pass an URL containing a `=` character into an environment variable for a `scalingo run` command and here's the error I encountered:

```
scalingo run -a pix-api-integration -e TEST=1=2 'echo $TEST'
 !     An error occurred:
       Invalid environment, format is '--env VARIABLE=value'
```

I think it should be possible to pass environment variables containing `=` in their value. It looks like the argument parser is a bit too restrictive here.

jonathanperret:
Fixes #673.

I took the liberty of adding a couple of tests for the `buildEnv` function. I don't write Go daily, so it's likely I did something suboptimally, please feel free to let me know.


EtienneM (approved):
Thanks a lot! We will try to schedule a release of the CLI soon

PR #675, opened 2021-08-24
fix(crontasks): display an empty table in case of 404

EtienneM:
Displaying an empty table when there is nothing is the current behaviour. See for instance the collaborators, autoscalers, alerts, etc. 

Related to [CRON-66]

- [ ] ~Add a changelog entry in the section "To Be Released" of CHANGELOG.md~: no need, this is still part of the upcoming cron product

[CRON-66]: https://scalingo.atlassian.net/browse/CRON-66

### DIFF
--- a/crontasks/list.go
+++ b/crontasks/list.go
@@ -1,21 +1,29 @@
 package crontasks
 
 import (
-	"github.com/Scalingo/cli/config"
-	"github.com/olekukonko/tablewriter"
-	"github.com/pkg/errors"
 	"os"
+
+	"github.com/Scalingo/cli/config"
+	httpclient "github.com/Scalingo/go-scalingo/v4/http"
+	"github.com/Scalingo/go-utils/errors"
+	"github.com/olekukonko/tablewriter"
+	"gopkg.in/errgo.v1"
 )
 
 func List(app string) error {
 	client, err := config.ScalingoClient()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


crontasks/list.go
```diff
@@ -1,21 +1,29 @@
...
 	if err != nil {
-		return errors.Wrap(err, "fail to get Scalingo client")
+		return errgo.Notef(err, "fail to get Scalingo client")
 	}
 
 	cronTasks, err := client.CronTasksGet(app)
 	if err != nil {
-		return errors.Wrap(err, "fail to get cron tasks")
+		rootError := errors.ErrgoRoot(err)
+		if !httpclient.IsRequestFailedError(rootError) || rootError.(*httpclient.RequestFailedError).Code != 404 {
```
curzolapierre, line 22: What happen if the `rootError` is not a `httpclient.RequestFailedError` ?
Does it throw an error because `.Code` cannot be accessible ?

curzolapierre (commented):
I have just a question, otherwise LGTM

crontasks/list.go
EtienneM: If `rootError` is not a `httpclient.RequestFailedError`, the condition evaluates to `true`. This is a _or_ condition, not a _and_.

curzolapierre: Oh, I read too quickly.. mb

curzolapierre (approved):
LGTM

PR #679, opened 2021-09-23
Implemented deploy --no-follow flag

### LINKED ISSUE #650: [deploy] an option should allow a user not to wait for the end of the deployment (just triggering it)
In some cases, where the second/minute is charged for instance, some users want the fastest pipeline possible and get the result of a deployment by a hook instead of synchronously.

A flag like `--no-follow` would allow them to only trigger a deployment and then finishing the command.

stopostit:
Fix #650

### DIFF
--- a/cmd/deployments.go
+++ b/cmd/deployments.go
@@ -111,6 +111,7 @@ var (
 		Usage:    "Trigger a deployment by archive",
 		Flags: []cli.Flag{appFlag,
 			cli.BoolFlag{Name: "war, w", Usage: "Specify that you want to deploy a WAR file"},
+			cli.BoolFlag{Name: "no-follow", Usage: "Return immediately after the deployment is triggered"},
 		},
 		Description: ` Trigger the deployment of a custom archive for your application
 
@@ -123,8 +124,11 @@ var (
 		$ scalingo -a myapp deploy archive.tar.gz v1.0.0
 		or
 		$ scalingo -a myapp deploy http://example.com/archive.tar.gz v1.0.0
+		or
+		$ scalingo --app my-app deploy --no-follow archive.tar.gz v1.0.0
+		$ scalingo --app my-app deployment-follow
 
-    # See also commands 'deployments'
+    # See also commands 'deployments, deployment-follow'
 `,
 		Action: func(c *cli.Context) {
 			args := c.Args()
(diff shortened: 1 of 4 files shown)


cmd/deployments.go
```diff
@@ -111,6 +111,7 @@ var (
 		Usage:    "Trigger a deployment by archive",
 		Flags: []cli.Flag{appFlag,
 			cli.BoolFlag{Name: "war, w", Usage: "Specify that you want to deploy a WAR file"},
+			cli.BoolFlag{Name: "no-follow", Usage: "Disable the display of deployment process if the first API call succeeded"},
```
EtienneM, line 114: I don't think the help message should talk about "API call". Our customers do not know the underlying logic of the `deploy` command, they don't know if there is some API calls or some other magic. What bother them is that when they call this command, it hangs for a long time, then displays the deployment logs, and eventually returns. Whereas there are some situations they want the command to return right after the deployment is triggered. What do you think about this wording:

```suggestion
			cli.BoolFlag{Name: "no-follow", Usage: "Return immediately after the deployment is triggered"},
```

EtienneM: I would also update the `Description` of this command to add an example with this flag. Something like:

```
$ scalingo --app my-app deploy --no-follow archive.tar.gz v1.0.0
$ scalingo --app my-app deployment-follow
```
The `deployment-follow` call is what you should call if you want to follow the deployment you just triggered.

There is a "See also commands" message in the description and I would add `'deployment-follow'` in it.

deployments/deploy.go
```diff
@@ -52,8 +57,12 @@ func Deploy(app, archivePath, gitRef string) error {
 
 	scalingoio.Status("Your deployment has been queued and is going to start…")
 
-	go showQueuedWarnings(c, app, deployment.ID)
+	if opts.NoFollow {
+		scalingoio.Status(fmt.Sprintf("The no-follow flag has been passed. you can check deployment logs with scalingo -a %s deployment-follow", app))
```
EtienneM, line 61: nitpick: 
- replace the use of `Status(fmt.Sprintf(` with `Statusf`
- in informative messages targeted to users, I prefer the use of long options. This is for example the case in our documentation: https://doc.scalingo.com/platform/app/logs
- (i'm not sure about this one.. 😅 ) I think "is passed" is a better conjugation than "has been passed".
- Missing upper case

```suggestion
		scalingoio.Statusf("The no-follow flag is passed. You can check deployment logs with scalingo --app %s deployment-follow", app)
```

EtienneM (commented):
Just adding some comments about the content of the documentation. Otherwise it looks overall good to me

EtienneM (changes requested):
Test

EtienneM (commented):
gregregre

EtienneM (approved):

EtienneM (changes requested):
Test

EtienneM (approved):
Nice! 👍🏾

Intrepidd:
Thank you so much for implementing this ! I'd love to use it ASAP, would it be possible to release it ? Thanks !

EtienneM:
> Thank you so much for implementing this ! I'd love to use it ASAP, would it be possible to release it ? Thanks !

Hi @Intrepidd! We plan to release a new version by the end of next week :)

Intrepidd:
Thanks so much ! On my way to reduce my deploy scripts by 90% 😅

PR #689, opened 2021-11-02
Feat(addons-info): Add addons-info command and documentation

### LINKED ISSUE #578: [Addons] Able to see addon informations
Add a command like `apps-info` but for addons, to be able to see if the addon is deployed, or being deployed and see some information about the addon.

For instance, I expect the command to be named `addons-info` and behave like:

```
oscalingo --app my-app addons-info --addon ad-XXX-1234
+-----------------------------+
| Database Type |  mongo |
| Version | 4.0.17 |
| Status | running |
| Plan | mongo-business-2048  | 
| Force TLS | true |
| Internet Accessibility | false |
```

sihamais:
Add `addons-info` command to display addon information  :
```bash
$ scalingo --app my-app addons-info --addon <addon-id>
+------------------------+--------------------+
| Database Type          | postgresql         |
| Version                | 13.3.0-2           |
| Status                 | running            |
| Plan                   | postgresql-sandbox |
| Force TLS              | activated          |
| Internet Accessibility | disabled           |
+------------------------+--------------------+
```
---

Fix #578


README.md
```diff
@@ -69,6 +69,7 @@ COMMANDS:
      addons-add        Provision an add-on for your application
      addons-remove     Remove an existing addon from your app
      addons-upgrade    Upgrade or downgrade an add-on attached to your app
+     addons-info       Display infos about add-ons attached to your app
```
EtienneM, line 72: The word "information" is uncountable in English. Moreover this call is about a single addon, hence I would remove the `s`
```suggestion
     addons-info       Display info about add-on attached to your app
```

cmd/addons.go
```diff
@@ -116,4 +117,32 @@ var (
 			autocomplete.AddonsUpgradeAutoComplete(c)
 		},
 	}
+	AddonsInfoCommand = cli.Command{
+		Name:     "addons-info",
+		Category: "Addons",
+		Usage:    "Display infos about add-ons attached to your app",
+		Flags:    []cli.Flag{appFlag, addonFlag},
+		Description: ` Display infos about add-ons attached to your app:
+    $ scalingo --app my-app addons-info --addon <addon-id>
```
EtienneM, line 126: Uncountable + use the full word
```suggestion
		Usage:    "Display information about an add-on attached to your app",
		Flags:    []cli.Flag{appFlag, addonFlag},
		Description: ` Display information about an add-on attached to your app:
    $ scalingo --app my-app addons-info --addon <addon-id>
```

EtienneM (changes requested):
The description of your PR should be `Fix #578`. With the keyword `Fix`, GitHub will automatically close the related issue when merging the PR

You forgot to update the CHANGELOG file

README.md
sihamais: I thought so too. But since all the others commands (addons-add, addons-remove, addons-upgrade...) have an `s` I got confused.

addons/info.go
```diff
@@ -0,0 +1,50 @@
...
+		return errgo.Notef(err, "fail to get Scalingo client")
+	}
+	addonInfo, err := c.AddonShow(app, addon)
+	if err != nil {
+		return errgo.Mask(err, errgo.Any)
+	}
+
+	dbInfo, err := db.Show(app,addon)
+	if err != nil {
+		return errgo.Mask(err, errgo.Any)
```
EtienneM, line 27: `errgo.Mask` is a deprecated way of handling the errors. Please use `errgo.Notef` (https://www.notion.so/scalingooriginal/Go-Style-Guide-de47c3b8d6c24c2e9b70c7b9c2f36660#158c056247cf463c9ae078893bf360b9)

addons/info.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	forceSsl, internetAccess := "disabled", "disabled"
+	if featuresLen := len(dbInfo.Features); featuresLen > 0 {
```
EtienneM, line 31: nitpick: we don't really use the construct you used as we don't find it easily readable.
```suggestion
	featuresLen := len(dbInfo.Features)
	if featuresLen > 0 {
```

addons/info.go
```diff
@@ -0,0 +1,50 @@
...
+		if featuresLen == 2 {
```
EtienneM, line 32: This is not really true. There is a possibility that length is greater than 2 and that internet access is enabled.

addons/info.go
```diff
@@ -0,0 +1,50 @@
...
+	t.Append([]string{"Status", fmt.Sprintf("%v", addonInfo.Status)})
+	t.Append([]string{"Plan", fmt.Sprintf("%v", addonInfo.Plan.Name)})
+	t.Append([]string{"Force TLS", forceSsl})
+	t.Append([]string{"Internet Accessibility", internetAccess})
+
+	t.Render()
+
+	return nil
+
+}
```
EtienneM, line 50: Add a blank line

EtienneM (changes requested):
Can you edit the description of the PR to include an example of execution so that I can see it?

addons/info.go
sihamais: Do you know where I can find the list of possible features for a db ? Thought they were only two.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### To be Released
 
+* feat(addons-info): add command `addons-info` to display information of an add-on [#578](https://github.com/Scalingo/cli/pull/689)
```
EtienneM, line 5:
```suggestion
* feat(addons-info): add command `addons-info` to display information of an add-on [#689](https://github.com/Scalingo/cli/pull/689)
```

addons/info.go
```diff
@@ -0,0 +1,52 @@
...
+		return errgo.Notef(err, "fail to get database information")
+	}
+
+	forceSsl, internetAccess := "disabled", "disabled"
+	for i := range dbInfo.Features {
+		if dbInfo.Features[i]["name"] == "force-ssl" {
+			forceSsl = strings.ToLower(dbInfo.Features[i]["status"])
+		} else if dbInfo.Features[i]["name"] == "publicly-available" {
+			internetAccess = strings.ToLower(dbInfo.Features[i]["status"])
+			break
```
EtienneM, line 37: You shouldn't break in this condition, there is no guarantee of the features order

```suggestion
```

EtienneM (changes requested):

addons/info.go
sihamais: Actually there is, since you cannot make your addon publicly available if SSL isn't enforced, so it has to be after the SSL entry. I've checked. What do you think ?

addons/info.go
EtienneM: We don't maintain a comprehensive list of features. Do you have access to appsdeck-database code?

addons/info.go
EtienneM: Even though you supposedly can't enforce Public availability without Force TLS, the order of the features in the array is not guaranteed at all in the code. I can imagine a scenario where this is not the case. For instance, in case of issue with these features, an operator can manually work on the database and the order of features on the database is not guaranteed

sihamais: I see, I'll remove it then. Thanks for the explanation.

addons/info.go
sihamais: I don't think so.

EtienneM: OK so you don't have access to the full list of features.. 😓

CHANGELOG.md
EtienneM: There is still this one

EtienneM (approved):

PR #693, opened 2021-11-03
feat(pgsql-console): add aliases

### LINKED ISSUE #676: [pgsql-console] Add some useful aliases
I always get confused about how this command is named. I would add two aliases: `psql-console` (`psql` is the name of the CLI tool) and `postgresql-console`

sihamais:

Fix #676

### DIFF
--- a/cmd/pgsql.go
+++ b/cmd/pgsql.go
@@ -10,6 +10,7 @@ import (
 var (
 	PgSQLConsoleCommand = cli.Command{
 		Name:     "pgsql-console",
+		Aliases: []string{"psql-console","postgresql-console"},
 		Category: "Databases",
 		Usage:    "Run an interactive console with your PostgreSQL addon",
 		Flags: []cli.Flag{appFlag,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### To be Released
 
-* feat(pgsql-console): add `psql-console` and `postgresql-console` for `pgsql-console` command []()
+* feat(pgsql-console): add `psql-console` and `postgresql-console` for `pgsql-console` command [693](https://github.com/Scalingo/cli/pull/693)
 * feat(open): add command `open` to open app on default browser [#691](https://github.com/Scalingo/cli/pull/691)
 * feat(addons-info): add command `addons-info` to display information of an add-on [#689](https://github.com/Scalingo/cli/pull/689)
 * feat(dashboard): add command `dashboard` to open dashboard of specified app on default browser [#686](https://github.com/Scalingo/cli/pull/686)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### To be Released
 
-* feat(pgsql-console): add `psql-console` and `postgresql-console` aliases for `pgsql-console` command [#693](https://github.com/Scalingo/cli/pull/693)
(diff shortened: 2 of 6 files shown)


cmd/pgsql.go
```diff
@@ -10,6 +10,7 @@ import (
 var (
 	PgSQLConsoleCommand = cli.Command{
 		Name:     "pgsql-console",
+		Aliases: []string{"psql-console","postgresql-console"},
```
EtienneM, line 13: Haha I didn't about this attribute, look at how I did it for another alias 😄 

https://github.com/Scalingo/cli/blob/5b11bd7de1a97f8c4c4ce70e98e8c023571af67d/cmd/deployments.go#L36-L44

What I prefer with my solution, is that the output of `scalingo help` does not mention this alias. It is some kind of typo helper and I don't really want to advertise it. 

And in this case my opinion I have mixed feeling whether I want to advertise about these aliases in `scalingo help`. Do you have an opinion on this?

sihamais: I truly don't have any opinion on this matter. As you like.

brandon-welsch: I agree with Etienne. In addition, it overloads the help command i think.

Afterwards, if we could display these aliases in the help of the command directly rather than in the `scalingo help`, that would be a plus, but that's a detail

curzolapierre: On my side I was confused the first time I saw the "Étienne's" solution in the `deployment-` autocomplete because it first didn't give me the impression that it was an alias but another command.
So I am interested in the other solution, maybe this one can be better.
Personally I find it cleaner this:
```
   Databases:                                                                                                          
     redis-console                                    Run an interactive console with your Redis addon                 
     mongo-console                                    Run an interactive console with your MongoDB addon               
     mysql-console                                    Run an interactive console with your MySQL addon                 
     pgsql-console, psql-console, postgresql-console  Run an interactive console with your PostgreSQL addon            
     influxdb-console                                 Run an interactive console with your InfluxDB addon              
```
instead of duplication of line like:
```
   Deployment:                                                                                                         
     deployments              List app deployments                                                                     
     deployment-logs          View deployment logs                                                                     
     deployment-follow        Follow deployment event stream                                                           
     deploy                   Trigger a deployment by archive                                                          
     deployment-delete-cache  Reset deployment cache                                                                   
     deployment-cache-delete  Reset deployment cache                                                                   
```

EtienneM: OK I think I eventually agree with Pierre an Léo. Let's use this `Aliases` attribute! Can you also update the `deployment-cache-delete` so that it's just an alias to `deployment-delete-cache`? Thanks 🙏🏾

EtienneM (changes requested):

cmd/pgsql.go
sihamais: Same for `self` and `whoami` ?
And should I keep the duplicated code or not (or maybe comment it) ?

EtienneM: Yes, please do the same for `self` and `woami`. Keep `self` and make `whoami` an alias of `self`.

You can remove the duplicated code.

sihamais: I'm on it.

sihamais: @curzolapierre
The alias inherits from it's original command for the autocompletion of the flags.

EtienneM (approved):

PR #692, opened 2021-11-03
feat(router-logs): add command to enable/disable router-logs

sihamais:
- [x] Add a changelog entry in the section "To Be Released" of CHANGELOG.md

Fix #605

Issue #605, opened 2020-11-30
Add a command to activate router logs


EtienneM:
The command could be named `router-logs`. Take example on the command `force-https`.

PR #701, opened 2021-11-05
fix: re-sync go-scalingo with the actual dependency (v4.15.1)

EtienneM:
The PR introducing the router logs feature to the CLI (#692) directly modified the vendor directly. This is not something we should do. This PR uses a new go-scalingo release (v4.15.0) which includes the function `AppsRouterLogs` introduced in the PR #692.

Related to #605


curzolapierre (approved):
LGTM

PR #705, opened 2021-11-10
fix(term): prefer the use of the os package rather than syscall

EtienneM:
The syscall package documentation states:

> Package syscall contains an interface to the low-level operating system primitives. The details vary depending on the underlying system. [...] The primary use of syscall is inside other packages that provide a more portable interface to the system, such as "os", "time" and "net". Use those packages rather than this one if you can.

https://pkg.go.dev/syscall

With this PR, the CLI compiles again on all OS

### DIFF
--- a/term/password.go
+++ b/term/password.go
@@ -2,7 +2,7 @@ package term
 
 import (
 	"fmt"
-	"syscall"
+	"os"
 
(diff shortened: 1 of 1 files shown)


curzolapierre (approved):
LGTM

PR #709, opened 2022-01-03
style(golangci-lint): autofixes

EtienneM:
Updated the CodeShip configuration according to our new procedure

### DIFF
--- a/apps/run/run_unix.go
+++ b/apps/run/run_unix.go
@@ -1,3 +1,4 @@
+//go:build darwin || dragonfly || freebsd || linux || netbsd || openbsd
 // +build darwin dragonfly freebsd linux netbsd openbsd
 
 package run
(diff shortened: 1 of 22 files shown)


apps/run/run_unix.go
```diff
@@ -1,3 +1,4 @@
+//go:build darwin || dragonfly || freebsd || linux || netbsd || openbsd
```
EtienneM, line 1: As specified in the [Build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) doc:

> Go versions 1.16 and earlier used a different syntax for build constraints, with a "// +build" prefix. The gofmt command will add an equivalent //go:build constraint when encountering the older syntax.

curzolapierre (approved):
LGTM

PR #728, opened 2022-06-03
chore(go): use go 1.17

EtienneM:
The fix for CVE-2022-29526 is only available in Go 1.17 and 1.18.


curzolapierre (approved):
Need to update the CI, otherwise LGTM

PR #731, opened 2022-06-13
fix(domains): replace DomainsUpdate with Domain*Certificate

EtienneM:
Fix https://github.com/Scalingo/go-scalingo/issues/249


domains/ssl.go
```diff
@@ -17,12 +17,12 @@ func DisableSSL(app string, domain string) error {
 
 	d, err := findDomain(c, app, domain)
 	if err != nil {
-		return errgo.Mask(err)
+		return errgo.Notef(err, "fail to find the matching domain to disable SSL")
 	}
 
-	_, err = c.DomainsUpdate(app, d.ID, "", "")
+	_, err = c.DomainUnsetCertificate(app, d.ID)
```
EtienneM, line 23: This is the actual fix of the related issue

PR #741, opened 2022-07-22
fix(autoscalers): min containers cannot be 1

### LINKED ISSUE #716: Autoscaler min-container example to 1 but must be at least 2
`
DESCRIPTION:
   Add an autoscaler to an application. It will automatically scale the application up or down depending on the target defined for the given metric.

   All options are mandatory.

   Example
     scalingo --app my-app autoscalers-add --container-type web --metric cpu --target 0.75 --min-containers 1 --max-containers 3

$ scalingo -a xxx autoscalers-add --container-type web --metric cpu --target 0.75 --min-containers 1 --max-containers 3
 !     An error occurred:
       * min_containers → must be at least set to 2
`

Minimum 2 is documented. But still hope you find a solution to come back to 1 !

EtienneM:
Fix #716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### To be Released
 
+* fix(autoscalers): min containers cannot be 1 [#741](https://github.com/Scalingo/cli/pull/741)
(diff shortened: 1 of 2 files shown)


brandon-welsch (approved):
LGTM. Just need to rebase for the changelog file

PR #748, opened 2022-07-22
fix(install.sh): better error message if fails to get the version

### LINKED ISSUE #714: [install.sh] Better errors management of download and installation of the CLI
I think we can add better management for the CLI download and installation. To detect certain errors in advance and thus be able to display a clearer message instead of doing commands after that a command has failed.

The first 2 examples below are linked if the customer have an old version of curl: https://github.com/Scalingo/cli-dl/issues/3#issuecomment-1017284152

### Version variable is empty

Version variable here can be empty if curl fail:
https://github.com/Scalingo/cli/blob/168b82e3e725f1b56276102e1ac752da08605f2f/dists/install.sh#L104

Afterwards we trying to extract the archive that doesn't exist, look the archive name, the version is missing:
```shell
-----> Extracting... Archive: /tmp/scalingo_cli_zZq/scalingo__darwin_amd64.zip
```

#### Fix

Stop the execution if we detect that the version is empty and display a message:
```
-----> Fail to get the version of the CLI.
You probably have an old version of curl. Please check your curl version and update accordingly.
```
or something like that
OR
If the version is empty use `latest` to fetch the latest version instead

### Download can fail

On this line the download can fail: https://github.com/Scalingo/cli/blob/168b82e3e725f1b56276102e1ac752da08605f2f/dists/install.sh#L110

Afterwards we trying to extract the archive that doesn't exist

#### Fix

Stop the execution if the download fail and display a message : "-----> Fail to download the CLI" or something like that.
And always displaying the curl error message.

### Extract can fail

On these lines the archive extract can fail: https://github.com/Scalingo/cli/blob/master/dists/install.sh#L113-L120

Afterwards we trying to install the binary that doesn't exist

#### Fix

Stop the execution (not try to install afterwards) if the extract command failed and display a message : "-----> Fail to extract the CLI binary" or something like that.

### Example of full log

```
% Total % Received % Xferd Average Speed Time Time Time Current
Dload Upload Total Spent Left Speed
100 3742 0 3742 0 0 3742 0 --:--:-- --:--:-- --:--:-- 11513
-----> Downloading Scalingo client... DONE
-----> Extracting... Archive: /tmp/scalingo_cli_zZq/scalingo__darwin_amd64.zip
End-of-central-directory signature not found. Either this file is not
a zipfile, or it constitutes one disk of a multi-part archive. In the
latter case the central directory and zipfile comment will be found on
the last disk(s) of this archive.
unzip: cannot find zipfile directory in one of /tmp/scalingo_cli_zZq/scalingo__darwin_amd64.zip or
/tmp/scalingo_cli_zZq/scalingo__darwin_amd64.zip.zip, and cannot find /tmp/scalingo_cli_zZq/scalingo__darwin_amd64.zip.ZIP, period.
DONE
install: line 129: /tmp/scalingo_cli_zZq/scalingo__darwin_amd64/scalingo: No such file or directory
/!\ Scalingo client is already installed (version 1.22.0)
Do you want to replace it with version ? [Y/n] y
-----> Install Scalingo client to /usr/local/bin
mv: /tmp/scalingo_cli_zZq/scalingo__darwin_amd64/scalingo: No such file or directory
! Fail to install Scalingo client (return 1)
```

EtienneM:
Fix #714

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### To be Released
 
+* fix(install.sh): better error message if fails to get the version [#748](https://github.com/Scalingo/cli/pull/748)
 * feat(logs): possibility to get the addon logs by using its type (e.g. Redis) [#745](https://github.com/Scalingo/cli/pull/745)
 * fix(error): interpret raw newline [#744](https://github.com/Scalingo/cli/pull/744)
 * fix(stats): display memory as IEC size [#742](https://github.com/Scalingo/cli/pull/742)
--- a/dists/install.sh
+++ b/dists/install.sh
@@ -104,7 +104,7 @@ main() {
 
   version=$(curl --silent https://cli-dl.scalingo.com/version | tr -d ' \t\n')
   if [ -z "$version" ]; then
-    echo "-----> Fail to get the version of the CLI." >&2
+    echo "-----> Fail to get the version of the CLI" >&2
     echo "You probably have an old version of curl. Please check your curl version and update accordingly." >&2
     exit 1
   fi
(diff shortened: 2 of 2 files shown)


brandon-welsch (approved):
LGTM. Thanks 🤗

PR #752, opened 2022-07-29
feat(goreleaser): use github actions to make release via goreleaser

### LINKED ISSUE #749: Investigate GoReleaser for the CLI releases
https://goreleaser.com/

curzolapierre:

You can take a look at https://github.com/curzolapierre/cli/releases/tag/1.23.11 for the result.

We need to test the install script once the release is done.

Fix #749


.goreleaser.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - arm64
+      - 386
+    ignore:
+      - goos: darwin
+        goarch: 386
+
+    # Custom ldflags templates.
+    # https://goreleaser.com/customization/templates/
+    ldflags:
+      - -X main.version={{.Version}}
```
curzolapierre, line 28: I wonder if we would prefer `Version` instead of `version`.
But I'm not sure why we are using this variable.

.github/workflows/publish.yml
```diff
@@ -1,43 +1,35 @@
...
-        run: |
-          echo "::set-output name=TAG_NAME::${GITHUB_REF#refs/tags/}"
-          echo "$TAG_NAME"
 
-      - name: Checkout the code
+    steps:
+      - name: Check out code into the Go module directory
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # See: https://goreleaser.com/ci/actions/
```
EtienneM, line 15: Noice 👍🏾

.github/workflows/publish.yml
```diff
@@ -1,43 +1,35 @@
...
 
       - name: Setup Go environment
         uses: actions/setup-go@v3
         with:
           go-version-file: "go.mod"
           check-latest: true
+        id: go
```
EtienneM, line 22: What is this `id` for?

.github/workflows/publish.yml
```diff
@@ -1,43 +1,35 @@
...
-        if: startsWith(github.ref, 'refs/tags/')
-        run: |
-          go install github.com/mitchellh/gox@latest
-          ./dists/make-release.sh -v ${{ steps.tag_name.outputs.TAG_NAME }}
-
-      - name: Release the new version
+      - name: Run GoReleaser
         if: startsWith(github.ref, 'refs/tags/')
-        uses: softprops/action-gh-release@v1
+        uses: goreleaser/goreleaser-action@master
```
EtienneM, line 30: I would prefer you use a version number here rather than `master`

```suggestion
        uses: goreleaser/goreleaser-action@v3
```

.goreleaser.yaml
```diff
@@ -0,0 +1,45 @@
+# Goreleaser documentation at https://goreleaser.com
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
```
EtienneM, line 4: nitpick: I don't see the purpose of this comment. We use Go modules everywhere
```suggestion
```

.goreleaser.yaml
```diff
@@ -0,0 +1,45 @@
+# Goreleaser documentation at https://goreleaser.com
```
EtienneM, line 1: nitpick: I don't see the purpose of this comment

```suggestion
```

.goreleaser.yaml
EtienneM: This line replaces the value of the variable `version` in the package `main`. This variable does not exist. I would say this line could be removed.

.goreleaser.yaml
```diff
@@ -0,0 +1,45 @@
...
+        -X main.githash={{.FullCommit}}
+
+archives:
+  -
+    name_template: 'scalingo_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}'
+    wrap_in_directory: true
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
```
EtienneM, line 39: After reading the doc about snapshot (https://goreleaser.com/customization/snapshots), I'm not sure I understand why we define a snapshot generation here

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### To be Released
 
+* feat(goreleaser): use goreleaser to make releases using github action [#749](https://github.com/Scalingo/cli/issues/749)
```
EtienneM, line 5: The changelog contains link to PR
```suggestion
* feat(goreleaser): use goreleaser to make releases using github action [#752](https://github.com/Scalingo/cli/issues/752)
```

EtienneM (changes requested):
If we go with GoReleaser, your PR should also remove all the previous scripts (`make-release`, `.goxc`, etc)

.github/workflows/publish.yml
curzolapierre: Don't needed I guess

EtienneM (approved):
🎉

PR #757, opened 2022-08-03
feat(workflows/publish): split in two jobs

EtienneM:
The `release` job is executed after `tests`, if it's successful.

Based on this doc: https://docs.github.com/en/actions/using-workflows/about-workflows#creating-dependent-jobs


curzolapierre (approved):
LGTM :+1:

PR #763, opened 2022-08-08
fix(installer) Mac OSX artefacts do not use zip anymore

johnsudaar:
Fix [SRE-688]

[SRE-688]: https://scalingo.atlassian.net/browse/SRE-688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dists/install.sh
+++ b/dists/install.sh
@@ -46,10 +46,9 @@ main() {
   uname -a | grep -qi 'Darwin' ; is_darwin=$?
 
   os=$(uname -s | tr '[A-Z]' '[a-z]')
-  ext=zip
+  ext='tar.gz'
   case $os in
     linux)
-      ext='tar.gz'
       ;;
     darwin)
       ;;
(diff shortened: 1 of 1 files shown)


dists/install.sh
```diff
@@ -46,10 +46,9 @@ main() {
   uname -a | grep -qi 'Darwin' ; is_darwin=$?
 
   os=$(uname -s | tr '[A-Z]' '[a-z]')
-  ext=zip
+  ext='tar.gz'
```
EtienneM, line 49: The switch case after this line could be simplified in a single if, don't you think?

dists/install.sh
```diff
@@ -122,14 +121,7 @@ main() {
...
   status "Extracting...   "
-  case $ext in
-    zip)
-      unzip -d "${tmpdir}" "${tmpdir}/${archive_name}"
-      ;;
-    tar.gz)
-      tar -C "${tmpdir}" -x -f "${tmpdir}/${archive_name}"
-      ;;
-  esac
+  tar -C "${tmpdir}" -x -f "${tmpdir}/${archive_name}"
```
EtienneM, line 124: Are we sure `tar` is available on all MacOS?

johnsudaar: I asked @ksol and it ran on his mac.

EtienneM (approved):

dists/install.sh
ksol: pretty sure it's been available for quite some time

Issue #574, opened 2020-06-25
Display changelog when upgrading the CLI


johnsudaar:
It might be a nice addition to display the changelog when we upgrade the CLI.

Something like:

```
$ curl -O https://cli-dl.scalingo.io/install && bash install
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3689    0  3689    0     0  23201      0 --:--:-- --:--:-- --:--:-- 23201
-----> Downloading Scalingo client...  DONE                  
-----> Extracting...   DONE                                                                                                                 
 /!\   Scalingo client is already installed (version 1.17.0           
client                                                                
                                                                      
)                                                                     
       Do you want to replace it with version 1.18.1? [Y/n] y         
-----> Install scalingo client to /usr/local/bin                      
       sudo required...                                               
[sudo] password for john:                                                                                                                   
-----> Installation completed, the command 'scalingo' is available. 
Here's what's new in this version:

Fix infinite recursion on some events with scalingo timeline #570
Fix log drain response parsing #570

You can find the entire changelog here: https://github.com/Scalingo/cli/blob/1.18.1/CHANGELOG.md
```

PR #767, opened 2022-08-12
Feature/574/add cli changelog command

curzolapierre:

Add a new command: `changelog` that shows the last release's changelog.
Now we show the last release's changelog when install the CLI.

## Example 

`changelog` command:
```
go build -o scalingo-cli ./scalingo && ./scalingo-cli changelog
Changelog of the version 1.24.1

## What's Changed
* docs(readme): add additional information related to the release by @curzolapierre in https://github.com/Scalingo/cli/pull/761
* chore(version): restore VERSION file by @curzolapierre in https://github.com/Scalingo/cli/pull/762
* fix(installer) Mac OSX artefacts do not use zip anymore by @johnsudaar in https://github.com/Scalingo/cli/pull/763
* fix(goreleaser) Generate zip archives for windows by @johnsudaar in https://github.com/Scalingo/cli/pull/769


**Full Changelog**: https://github.com/Scalingo/cli/compare/1.24.0...1.24.1
```

Regarding the update of the `install.sh` command, it means we need to release a new CLI version right after this PR is merged so that we correctly output the changelog after an installation.

Fix https://github.com/Scalingo/cli/issues/574


EtienneM:
@curzolapierre Is this PR reviewable?

curzolapierre:
@EtienneM I think it is reviewable now.
I was wondering if we shouldn't do the [check](https://github.com/Scalingo/cli/blob/dcf3826c5bc33f3e18615c7824be1c2089cac994/update/check.go#L79) before executing the user's command.

Because currently, as we can see in the description on this PR, the changelog is shown after the execution of the PR, WDYT about this?

cmd/cli_changelog.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"github.com/urfave/cli"
+
+	"github.com/Scalingo/cli/cmd/autocomplete"
+	"github.com/Scalingo/cli/update"
+)
+
+var (
+	cliChangelogCommand = cli.Command{
+		Name:     "cli-changelog",
```
EtienneM, line 12: nitpick: `cli-changelog` seems a bit clumsy. What do you think about just naming it `changelog`?

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/dustin/go-humanize v1.0.0
 	github.com/fatih/color v1.13.0
 	github.com/go-git/go-git/v5 v5.4.2
+	github.com/google/go-github/v45 v45.2.0
```
EtienneM, line 16: v47 has been released if you could upgrade it

services/github/client.go
```diff
@@ -0,0 +1,37 @@
...
+	repo  = "cli"
+)
+
+type client struct {
+	githubRepoService *github.RepositoriesService
+}
+
+func (c client) ListReleases(ctx context.Context, opts *github.ListOptions) ([]*github.RepositoryRelease, error) {
+	repoReleases, githubResponse, error := c.githubRepoService.ListReleases(ctx, owner, repo, opts)
+	debug.Printf("github response: %#v\n", githubResponse)
```
EtienneM, line 22: nitpick
```suggestion
	debug.Printf("GitHub response: %#v\n", githubResponse)
```

services/github/client.go
```diff
@@ -0,0 +1,37 @@
...
+	return repoReleases, error
```
EtienneM, line 23: You should annotate the error and rename it to `err`, like we do everywhere 😅

services/github/client.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (c client) GetLatestRelease(ctx context.Context) (*github.RepositoryRelease, error) {
+	repoRelease, githubResponse, error := c.githubRepoService.GetLatestRelease(ctx, owner, repo)
+	debug.Printf("github response: %#v\n", githubResponse)
+	return repoRelease, error
```
EtienneM, line 29: You should annotate the error and rename it to `err`

services/github/client.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func New() client {
```
EtienneM, line 32: A public function shouldn't return a private structure. We usually return the interface.

suggestion: I find `NewClient` a better name

update/changelog.go
```diff
@@ -0,0 +1,72 @@
+package update
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"gopkg.in/errgo.v1"
+
+	githubClient "github.com/google/go-github/v45/github"
```
EtienneM, line 10:
```suggestion
	"gopkg.in/errgo.v1"
	githubClient "github.com/google/go-github/v45/github"
```

update/changelog.go
```diff
@@ -0,0 +1,72 @@
...
+		// Detect the release from the cache, we can show changelog from next release.
+		if cacheVersion == release.GetTagName() {
+			beginToSowRelease = true
+		}
+	}
+
+	return nil
+}
+
+// getLastReleases get last 10 releases.
```
EtienneM, line 57: Is there a reason to limit to 10 reasons? It would be interesting to explain this limit.

update/changelog.go
```diff
@@ -0,0 +1,72 @@
...
+	cliLastRelease, err := client.GetLatestRelease(ctx)
+	if err != nil {
+		return errgo.Notef(err, "fail to get last CLI release")
+	}
+	printRelease(*cliLastRelease)
+
+	return nil
+}
+
+func ShowChangelog(cacheVersion, newVersion string) error {
```
EtienneM, line 34: Can you add a comment to explain what it does? I don't completely understand the use case. The related issue suggests to display the last version changelog after installing it.

Moreover it should be private.

EtienneM (changes requested):
I think I like what the related issue suggests. i.e. printing the changelog after the CLI installation. I'm not keen on having it after the output of the command executed by the client. Moreover it would simplify the content of this PR.

services/github/client.go
curzolapierre: oops :sweat_smile:

services/github/client.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	owner = "scalingo"
+	repo  = "cli"
+)
+
+type Client struct {
+	githubRepoService *github.RepositoriesService
+}
+
+func (c Client) ListReleases(ctx context.Context, opts *github.ListOptions) ([]*github.RepositoryRelease, error) {
```
EtienneM, line 20: It seems this function is never used.

dists/install.sh
```diff
@@ -169,6 +169,7 @@ main() {
     error "Fail to install Scalingo client (return $rc)\n"
   else
     status "Installation completed, the command 'scalingo' is available.\n"
+    status "Here's what's new in this version:$(echo \\n\\n)$(curl https://api.github.com/repos/scalingo/cli/releases/tags/${version} | grep '"body": ' | cut -d ':' -f2- | cut -d '"' -f2- | sed 's/\\r\\n/\n/g')$(echo \\n\\n)"
```
EtienneM, line 172: It might be easier to simply call `scalingo changelog`, don't you think?

services/github/client.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (c Client) GetLatestRelease(ctx context.Context) (*github.RepositoryRelease, error) {
+	repoRelease, githubResponse, err := c.githubRepoService.GetLatestRelease(ctx, owner, repo)
+	debug.Printf("github response: %#v\n", githubResponse)
+	return repoRelease, err
+}
+
+func NewClient() Client {
+	githubClient := github.NewClient(nil)
```
EtienneM, line 33: We should never use the default HTTP client

update/changelog.go
```diff
@@ -0,0 +1,31 @@
...
+
+	githubClient "github.com/google/go-github/v47/github"
+	"gopkg.in/errgo.v1"
+
+	"github.com/Scalingo/cli/services/github"
+)
+
+func printRelease(release githubClient.RepositoryRelease) {
+	fmt.Printf("Changelog of the version %v\n\n", release.GetTagName())
+	fmt.Printf("%v\n\n", strings.ReplaceAll(*release.Body, "\\r\\n", "\r\n"))
```
johnsudaar, line 16: Should we check if Body is nil ?

update/changelog.go
```diff
@@ -0,0 +1,31 @@
...
+func ShowLastChangelog() error {
+	ctx := context.Background()
+
+	client := github.NewClient()
+
+	cliLastRelease, err := client.GetLatestRelease(ctx)
+	if err != nil {
+		return errgo.Notef(err, "fail to get last CLI release")
+	}
+	printRelease(*cliLastRelease)
```
johnsudaar, line 28: Should we check if cliLastRelease is nil ?

EtienneM: `cliLastRelease` is always defined if there is no error. So it should be safe :)

johnsudaar: It might break in an update but ok it could be nitpicking :)

services/github/client.go
```diff
@@ -0,0 +1,41 @@
...
+	return client{
+		githubRepositoriesService: github.NewClient(&http.Client{
+			Timeout: 5 * time.Second,
+		}).Repositories,
+	}
+}
+
+func (c client) GetLatestRelease(ctx context.Context) (*github.RepositoryRelease, error) {
+	latestRelease, githubResponse, err := c.githubRepositoriesService.GetLatestRelease(ctx, "Scalingo", "cli")
+	defer githubResponse.Body.Close()
```
johnsudaar, line 32: Should we check if githubResponse and Body could be nil ?

johnsudaar (approved):
LGTM

PR #776, opened 2022-08-31
feat(stacks): Add possibility to show deprecated stacks

aurelien-reeves-scalingo:
Refs. https://github.com/scalingo/api/issues/1831

### DIFF
--- a/stacks/list.go
+++ b/stacks/list.go
@@ -48,7 +48,7 @@ func List(ctx context.Context, withDeprecated bool) error {
 			}
 
 			t.Append([]string{stack.ID, stack.Name, stack.Description, defaultText, deprecatedText, deprecationDate})
-		} else {
+		} else if !stack.IsDeprecated() {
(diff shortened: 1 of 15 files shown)


stacks/list.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/Scalingo/cli/config"
 )
 
-func List(ctx context.Context) error {
+func List(ctx context.Context, withDeprecated bool) error {
```
curzolapierre, line 13: Nitpick: I don't find the name `withDeprecated` very representative for this variable.
Maybe something like `isWithDepractedFlag` better, WDYT?

curzolapierre (commented):
Just a picky comment, otherwise LGTM

stacks/list.go
aurelien-reeves-scalingo: I'll do the change

Issue #780, opened 2022-09-02
CurrentApp fails if using `--app` as a global option with urfave/cli v2


aurelien-reeves-scalingo:
With urfave/cli v2, the following command does not work:

```console
$ scalingo --app biniou restart
Unable to find the application name, please use --app flag.
exit status 1
```

It has to be rewritten as the following:
```console
$scalingo restart --app biniou
 !     An error occurred:
       * app → is new
exit status 1
```

It appears that https://github.com/Scalingo/cli/blob/master/detect/app.go#L19 is generating that error message

aurelien-reeves-scalingo:
Refs. https://github.com/Scalingo/cli/pull/781#issuecomment-1235312055

> The issue may be due to default values
> 
> There may be other flags to take care of. On a first glance, the `addon` one may be impacted as it use a similar approach.

PR #781, opened 2022-09-02
fix(cli): Fix detection --app, --region, --remote and --addon from command line arguments

aurelien-reeves-scalingo:
Fix https://github.com/Scalingo/cli/issues/780

- [x] Update CHANGELOG.md

### DIFF
--- a/detect/app.go
+++ b/detect/app.go
@@ -65,8 +65,8 @@ func GetAppNameFromGitRemote(directory string, remoteName string) (string, error
 
 func extractAppNameFromCommandLine(c *cli.Context) string {
 	for _, cliContext := range c.Lineage() {
-		if context.String("app") != "<name>" {
-			return context.String("app")
+		if cliContext.String("app") != "<name>" {
+			return cliContext.String("app")
 		}
 	}
 
--- a/detect/app.go
+++ b/detect/app.go
@@ -17,19 +17,22 @@ import (
 // app flag, environment variable "SCALINGO_APP", current Git remote
 // returns an empty string if not found
 func CurrentApp(c *cli.Context) string {
-	var appName string
 	var err error
 
-	if c.String("app") != "<name>" {
-		appName = c.String("app")
-	} else if os.Getenv("SCALINGO_APP") != "" {
-		appName = os.Getenv("SCALINGO_APP")
-	} else if dir, ok := utils.DetectGit(); ok {
-		appName, err = GetAppNameFromGitRemote(dir, c.String("remote"))
-		if err != nil {
-			debug.Println(err)
+	appName := extractAppNameFromCommandLine(c)
+	if appName == "<name>" {
+		appName = ""
+
+		if os.Getenv("SCALINGO_APP") != "" {
+			appName = os.Getenv("SCALINGO_APP")
+		} else if dir, ok := utils.DetectGit(); ok {
+			appName, err = GetAppNameFromGitRemote(dir, c.String("remote"))
+			if err != nil {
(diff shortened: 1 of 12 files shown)


aurelien-reeves-scalingo:
The issue may be due to default values

There may be other flags to take care of. On a first glance, the `addon` one may be impacted as it use a similar approach.

detect/app.go
```diff
@@ -60,6 +63,16 @@ func GetAppNameFromGitRemote(directory string, remoteName string) (string, error
 	return "", errgo.Newf("[detect] Scalingo Git remote hasn't been found")
 }
 
+func extractAppNameFromCommandLine(c *cli.Context) string {
+	for _, context := range c.Lineage() {
```
EtienneM, line 67: `context` is not a very good variable name as it is often the name of an imported package. Suggestion:

```suggestion
	for _, cliContext := range c.Lineage() {
```

EtienneM, line 67: I'm not sure I understand what does `Lineage` do (https://pkg.go.dev/github.com/urfave/cli/v2#Context.Lineage), but I trust you if you tested it and it fixes the problem ^^

Shouldn't we also use it for the other global flags such as the region? I think it has the same problem, don't you think?

EtienneM (changes requested):
nitpick:
- missing changelog
- PR is in draft

detect/app.go
aurelien-reeves-scalingo: There are two "app" flags defined: one global option, and one local to the actual command, both with default value "<name>".

`c.String("name")` in the context of a command returns the value for the local flag. When it is not set in the command line, it is the default value.

`c.Lineage()` returns an array with the local context, and all the parent contexts before, thus the root one with the global options.

So we can loop through that array to look for an `app` flag which has actually been set

Regarding the other `global` flags, indeed we have to investigate. I think the `addon` may be imacted as it uses a similar approach as the `app` one. For the other ones, more investigation are required

aurelien-reeves-scalingo:
Same issue with `--addon`. Working on it

aurelien-reeves-scalingo:
`--addon`: done

The `--remote` flag should be impacted too. However that flag is used only with `create` and `git-setup`. What about removing it from the global options? @EtienneM

aurelien-reeves-scalingo:
Changelog has been updated not with the addition of an entry, but with the removal of a previous breaking change notice

aurelien-reeves-scalingo:
> The `--remote` flag should be impacted too. However that flag is used only with `create` and `git-setup`, and no documentation refers to it as a global option. What about removing it from the global options? @EtienneM

After discussion with @Soulou, it is actually used with the app detection based on the remote.

```console
# The following command will detect the application name based on the given remote
$ scalingo --remote scalingo ps
```

So, keep the support for it, and still to be fixed with urfave/cli v2

detect/app.go
```diff
@@ -63,10 +63,21 @@ func GetAppNameFromGitRemote(directory string, remoteName string) (string, error
 	return "", errgo.Newf("[detect] Scalingo Git remote hasn't been found")
 }
 
+// Returns the remote name specified in command flags
```
EtienneM, line 66: Just a picky comment: in Go, function comments start with the function name

```suggestion
// RemoteNameFromFlags returns the remote name specified in command flags
```

This is for a prettier output in the Go doc. For example: https://pkg.go.dev/github.com/Scalingo/go-scalingo/v5#NewStaticTokenGenerator

EtienneM (approved):
Just a remark but I already validate it :)

PR #798, opened 2022-09-27
WIP: filter for all one-off

btrd:
fix https://github.com/Scalingo/cli/issues/799

### DIFF
--- a/apps/logs.go
+++ b/apps/logs.go
@@ -80,7 +80,7 @@ func checkFilter(ctx context.Context, c *scalingo.Client, appName string, filter
 		return nil
 	}
 
-	if filter == "router" {
+	if filter == "router" || filter == "one-off" {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


btrd:
From my local test, it doesn't work, maybe the [log-service](https://github.com/Scalingo/logs-service/) doesn't allow this filter ?

Soulou:
I confirm, this filter is not allowed, why:

It's filtered by the container_types

EtienneM (commented):
Currently our backend does not support this syntax. Is there a reason that the customer don't use `grep`? I mean, it should be possible to do:

```
scalingo --app my-app logs | grep one-off
```
WDYT?

btrd:
- to prevent from having to download every log lines
- grep doesn't work when using -f
- if `web` & `web-1` if valid, there is no reason (from the point of view of the client) that `one-off-1` work but not `one-off`

I think we can close this PR, I let you decide what to do with the issue.

btrd:
Need a change in https://github.com/Scalingo/logs-service to work.

PR #805, opened 2022-09-28
feature(data_access_consent) Add support for data_access_consent

johnsudaar:
This is the bare minimum, waiting for the full API to be released


Soulou (commented):
What's annoying me here is that there is plenty of commands which don't need consent but in this PR it seems everything requires consent.

- App Logs
- App Environment
- App One-off
- Database consoles
- Database db-tunnel

And that's it. The rest is platform configuration which is not submitted to explicit consent, only customer's data by themselves

Soulou (approved):
Can we make a type of beta release to distribute this internally before release?

Issue #735, opened 2022-06-23
[windows] 1.23.0 version of the CLI doesn't work properly


brandon-welsch:
```
Les commandes, par exemple “scalingo apps” ne retournent rien, tournent dans le vide sans rien afficher.
```

Impact a customer: https://app.intercom.com/a/inbox/w4oogu7s/inbox/shared/all/conversation/12375700072124

brandon-welsch:
On Windows with a Git Bash shell:
```
Brandon@Scalingo MINGW64 ~
$ scalingo apps
interrupt catched, aborting…

Brandon@Scalingo MINGW64 ~
$ DEBUG=1 scalingo apps
[DEBUG]2022/06/23 18:22:21 [Regions] Use the cache
```

Works with the `--region` flag.

johnsudaar:
Related to [SRE-573]

EtienneM:
It seems that it's the region detection which fails. The following does *not* work (hangs foreever):

```
DEUBG=1 scalingo --app my-app apps-info
```

But this work:

```
DEUBG=1 scalingo --region osc-fr1 --app my-app apps-info
```

brandon-welsch:
Another customer has the issue: https://app.intercom.com/a/apps/w4oogu7s/inbox/inbox/3379738/conversations/12375700081105

Issue #734, opened 2022-06-20
[windows] The CLI doesn't work at all on Windows with command that require an input from the user


brandon-welsch:
Log:
```bash
$ scalingo login
        Trying login with SSH
 !      SSH Connection failed.
        Trying login with user/password:

        Username or email: [REDACTED]
        Password:   ! Fail to login (1/3): fail to read the password on stdin: The handle is invalid

        Username or email: interrupt catched, abording
```
And the command doesn't display anything with `mysql-console` for example.

Seems related to https://github.com/99designs/aws-vault/issues/772

Need to test this fix: https://github.com/99designs/aws-vault/pull/113

A customer had this issue: https://app.intercom.com/a/apps/w4oogu7s/inbox/inbox/3379738/conversations/12375700071636
Contact them when this is fixed

brandon-welsch:
From Windows with the Git Bash shell:
```
Brandon@Scalingo MINGW64 ~
$ scalingo.exe login
       Trying login with SSH…
 !     SSH connection failed.
       Trying login with user/password:

       Username or email: [REDACTED]
       Password:  !     Fail to login (1/3): fail to read the password on stdin: Descripteur non valide

       Username or email:
       Username or email:
       Username or email: interrupt catched, aborting…
```

johnsudaar:
Related to [SRE-573]

EtienneM:
Another impacted customer: https://app.intercom.com/a/inbox/w4oogu7s/inbox/shared/all/conversation/12375700080025

EtienneM:
This will be fixed for the next release including the fix in https://github.com/Scalingo/cli/pull/806. Probably 1.26.0.

EtienneM:
FYI the fix from https://github.com/99designs/aws-vault/pull/113 has been implemented in the PR #705:

https://github.com/Scalingo/cli/pull/705/files#diff-40837a1735f9474e15525ba28f225c256a9ab77d7c5f1fce60caa79e7a33243bR13

PR #806, opened 2022-09-30
feat(publish): compile Windows release on Windows

EtienneM:
With this PR, the `publish` action has two different jobs for the GoReleaser execution:

- `releases-windows`: execute GoReleaser on a Windows server so that we don't use Go cross compile feature. This is to prevent issues such as the one described in #734.
- `releases`: execute GoReleaser on a Ubuntu server for OpenBSD, Linux and macOS using Go cross compile feature. 

All these features are added as assets of a single GitHub release. You can see an example of execution on my fork: https://github.com/EtienneM/cli/releases/tag/1.25.2-beta8. One of the cons from my perspective is that we have two different files for the checksums. 

Related to #734 and #735

- [ ] ~Add a changelog entry in the section "To Be Released" of CHANGELOG.md~ no changelog entry, this is an internal change

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,7 +1,6 @@
 name: Test and Release Go CLI
 
-on:
-  push:
+on: [push]
 
 jobs:
   tests:
@@ -13,13 +12,13 @@ jobs:
         with:
           go-version-file: "go.mod"
           check-latest: true
-      - run: go version
+          cache: true
       - name: Execute the tests
         run: go test -race ./...
 
-  release:
+  releases:
     needs: tests
-    name: GoReleaser Build
+    name: GoReleaser Build on All OS but Windows
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
     steps:
@@ -30,7 +29,7 @@ jobs:
         with:
           go-version-file: "go.mod"
           check-latest: true
-      - run: go version
+          cache: true
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v3
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -1,7 +1,6 @@
 name: Test and Release Go CLI
 
-on:
-  push:
+on: [push]
```
EtienneM, line 3: This is just a way to write the same things with a different syntax, and I prefer this one.

.github/workflows/publish.yml
```diff
@@ -30,11 +29,36 @@ jobs:
...
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v3
+        with:
+          version: latest
+          args: release --config .goreleaser-windows.yaml --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: List files in the repository
+        run: |
+          ls ${{ github.workspace }}/dist
```
curzolapierre, line 64: Is this debug?

.github/workflows/publish.yml
```diff
@@ -30,11 +29,36 @@ jobs:
...
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # See: https://goreleaser.com/ci/actions/
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: "go.mod"
+          check-latest: true
+          cache: true
```
curzolapierre, line 54: Out of curiosity, what is the use of this `cache` property?

curzolapierre (changes requested):
I think, `./.goreleaser.yaml` file needs some modification to remove windows OS support.

.github/workflows/publish.yml
EtienneM: oups, yes

.github/workflows/publish.yml
EtienneM: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

> The action has a built-in functionality for caching and restoring go modules and build outputs.

It should speed up a bit the CI execution.

curzolapierre (approved):
LGTM

PR #835, opened 2022-12-23
fix(commands): display help usage message if needed

### LINKED ISSUE #831: [db-tunnel] If we use flags after the `DATABASE_URL` field the command will silently fail
For example:

```
scalingo -a john-test db-tunnel SCALINGO_POSTGRESQL_URL -p 10010
```
Will silently fail and just exit.

It would be nice to display a nice error message instead letting the client know that there is an issue.

### LINKED ISSUE #834: CLI does not show help of command when number of args is not correct anymore
## To reproduce:

```bash
scalingo --app my-app addons-add
```
This previous command does not shown anything

## Expected

To have the help of the command shown like
```bash
➜ scalingo --app my-app addons-add --help
NAME:
   scalingo addons-add - Provision an add-on for your application

USAGE:
   scalingo addons-add [command options] [arguments...]

CATEGORY:
   Addons

DESCRIPTION:
    Provision an add-on for your application:
       $ scalingo -a myapp addons-add <addon-name> <plan>

       # See also 'addons-list' and 'addons-plans'


OPTIONS:
   --app value, -a value  Name of the current app (default: "<name>")
   --region value         Name of the region to use
   --help, -h             show help (default: false)
```

EtienneM:
You may want to review commit by commit as this PR also contains a bunch of refactoring.

I think the scenario described in the related issue actually never worked.. 😅  

Fix #834

This PR also fixes #831

I would like to release a new version after this PR.

### DIFF
--- a/cmd/addons.go
+++ b/cmd/addons.go
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	AddonsListCommand = cli.Command{
+	addonsListCommand = cli.Command{
 		Name:     "addons",
 		Category: "Addons",
 		Usage:    "List used add-ons",
@@ -37,7 +37,7 @@ var (
 			autocomplete.CmdFlagsAutoComplete(c, "addons")
 		},
 	}
-	AddonsAddCommand = cli.Command{
+	addonsAddCommand = cli.Command{
 		Name:     "addons-add",
 		Category: "Addons",
 		Flags:    []cli.Flag{&appFlag},
@@ -65,7 +65,7 @@ var (
 			autocomplete.AddonsAddAutoComplete(c)
 		},
 	}
-	AddonsRemoveCommand = cli.Command{
+	addonsRemoveCommand = cli.Command{
 		Name:     "addons-remove",
 		Category: "Addons",
 		Flags:    []cli.Flag{&appFlag},
@@ -93,7 +93,7 @@ var (
 			autocomplete.AddonsRemoveAutoComplete(c)
 		},
 	}
-	AddonsUpgradeCommand = cli.Command{
+	addonsUpgradeCommand = cli.Command{
 		Name:     "addons-upgrade",
 		Category: "Addons",
 		Flags:    []cli.Flag{&appFlag},
@@ -121,7 +121,7 @@ var (
 			autocomplete.AddonsUpgradeAutoComplete(c)
 		},
 	}
-	AddonsInfoCommand = cli.Command{
+	addonsInfoCommand = cli.Command{
 		Name:     "addons-info",
 		Category: "Addons",
 		Usage:    "Display information about an add-on attached to your app",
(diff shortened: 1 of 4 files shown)


cmd/commands.go
```diff
@@ -63,9 +62,8 @@ func (cmds *AppCommands) addCommand(cmd Command) {
 			if currentRegion != "" {
 				config.C.ScalingoRegion = currentRegion
 			}
-			action(c)
 
-			return nil
+			return action(c)
```
EtienneM, line 66: With this change, if an action returns an error, it displays the error message (and that's it). There is no other messages than the error string representation.

cmd/addons.go
```diff
@@ -21,23 +21,22 @@ var (
 `,
 		Action: func(c *cli.Context) error {
 			currentApp := detect.CurrentApp(c)
-			var err error
-			if c.Args().Len() == 0 {
-				err = addons.List(c.Context, currentApp)
-			} else {
-				cli.ShowCommandHelp(c, "addons")
+			if c.Args().Len() > 0 {
+				return cli.ShowCommandHelp(c, "addons")
```
EtienneM, line 25: With the change explained here (https://github.com/Scalingo/cli/pull/835/files#r1056545792), this leads to the following output:

```
 ✘ emichon@biniou  ~/Documents/Scalingo/golang/src/github.com/Scalingo/cli   fix/834/addons-add_help  go build -o scalingo-cli ./scalingo && ./scalingo-cli --app biniou addons toto
No help topic for 'addons'
```

cmd/commands.go
```diff
@@ -26,49 +25,59 @@ type Command struct {
...
+		cmds.commands = append(cmds.commands, cmd.Command)
+		return
+	}
+
+	// Regional commands are modified before being added to the list of commands.
+	regionFlag := &cli.StringFlag{Name: "region", Value: "", Usage: "Name of the region to use"}
+	cmd.Command.Flags = append(cmd.Command.Flags, regionFlag)
+	// I don't really understand the official documentation about this field.
+	// But setting it to true enables the display of the help usage message when `cli.ShowCommandHelp` is called. Without this setting, the call to `cli.ShowCommandHelp` displays a "No help topic for command" error message.
+	cmd.Command.HideHelp = true
```
EtienneM, line 39: This is the actual fix of the related issue.

cmd/commands.go
```diff
@@ -26,49 +25,59 @@ type Command struct {
...
+			debug.Printf("[Regions] Use the default region '%s'\n", region.Name)
+			currentRegion = region.Name
+		}
 
-			return nil
+		if currentRegion != "" {
+			config.C.ScalingoRegion = currentRegion
 		}
+
+		return action(c)
```
EtienneM, line 79: This is a noticeable change. 

With this change, if an action returns an error, it displays the error message (and that's it). There is no other messages than the error string representation.

Before this PR, if an action returned an error it would simply be ignored.

curzolapierre (approved):
Thanks, LGTM

PR #863, opened 2023-01-06
fix(cli-build) Compile the CLI statically to prevent GLIBC incompatibility

### LINKED ISSUE #862: Last version may fail to run with `version 'GLIBC_2.32' not found`
From @johnsudaar:

> I'm wondering, is there a reason why we don't build with `CGO_ENABLED=0`? It sounds like a good idea to statically compile our CLI WDYT?
> We're using ubuntu-latest so this might've changed. We also changed the go-releaser version.

Steps to reproduce:

```
docker run --pull=always --rm -it ruby:3.1.3 bash
curl -O https://cli-dl.scalingo.com/install && bash install
```

There is already n issue during  the installation:

```
-----> Installation completed, the command 'scalingo' is available.
scalingo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by scalingo)
scalingo: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by scalingo)
-----> Here's what's new in this version:


```

johnsudaar:
Fix #862

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -37,6 +37,7 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CGO_ENABLED: 0
 
   releases-windows:
     needs: tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### To be Released
 
+* fix(cli-build) Compile the CLI statically to prevent GLIBC incompatibility [#863](https://github.com/Scalingo/cli/pull/863)
+
 ### 1.27.0
 
 * fix(commands): display help usage message if needed [#835](https://github.com/Scalingo/cli/pull/835)


.github/workflows/publish.yml
```diff
@@ -37,6 +37,7 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CGO_ENABLED: 0
```
EtienneM, line 40: Isn't it also important to add this flag for the Windows version?

johnsudaar: I'm not sure and since i do not have a way to test it. I'd like to be safe WDYT ?

EtienneM (approved):

.github/workflows/publish.yml
johnsudaar: Running the build on my fork to ensure that it works: https://github.com/johnsudaar/cli/actions/runs/3855819102

Issue #865, opened 2023-01-09
`--help` option does not work anymore


EtienneM:
Following the PR  #835, the `--help` option does not work anymore:

```
 scalingo --app biniou integrations-add --help
Incorrect Usage: flag: help requested

Fail to run command: flag: help requested
```

Workaround in the meantime is to use the `help` command:

```
 scalingo --app biniou help integrations-add  
NAME:
   scalingo integrations-add - Link your Scalingo account with your account on a tool such as github.com

USAGE:
   scalingo integrations-add [command options] [arguments...]

CATEGORY:
   Integrations

DESCRIPTION:
   Link your Scalingo account with your account on a tool. After creating the link, you will be able to automatically deploy when pushing to your repository, or create Review Apps for all pull requests created.

     For github.com:
     $ scalingo integrations-add github

     For gitlab.com:
     $ scalingo integrations-add gitlab

     For GitHub Enterprise:
     $ scalingo integrations-add --url https://ghe.example.com --token personal-access-token github-enterprise

     For GitLab Self-hosted:
     $ scalingo integrations-add --url https://gitlab.example.com --token personal-access-token gitlab-self-hosted

     # See also commands 'integrations', 'integrations-delete', 'integrations-import-keys'

OPTIONS:
   --app value, -a value  Name of the current app
   --url value            URL of the integration
   --token value          Token of the integration
   --help, -h             show help (default: false)
```

PR #868, opened 2023-01-20
fix(commands): cli help refactoring

### LINKED ISSUE #869: Argument usage not enough documented
In CLI utilities, when you want to know more about the available arguments, it shall be displayed in the command usage. For example : 

```
➜  ~ scalingo help create
NAME:
   scalingo create - Create a new app

USAGE:
   scalingo create [command options] [arguments...]
```

Here we have a placeholder `[arguments...]` even when there's no arguments waited.

We need no read all the documentation and examples in order to find if there's an awaited argument.

This command in fact waits for a mandatory argument : the application name. The correct usage should be the following :

```
➜  ~ scalingo help create
NAME:
   scalingo create - Create a new app

USAGE:
   scalingo create [command options] app-name
```

So this would be clearer in the CLI help.

### LINKED ISSUE #870: Inconsistent command description
In the CLI help, we have pretty inconsistent messages formatting. 

Sometimes we will have an usage which duplicates the USAGE displayed on top. 
Sometimes we have examples which aren't formatted the same. Sometimes three spaces, sometimes a `$` prefixing. Sometimes there are one or two spaces before the message description.

We should add consistency to the help description.

yohann-bacha:
Fix #869 
Fix #870 

This PR tries to bring coherence to scalingo cli help messages.

This pull request brings :
- Arguments needed (or optional) in the usage of the command. Before, we only had `[arguments...]` whether we had args or not. Now, available arguments are displayed in the usage, and when there's not, the `[arguments...]` placeholder does not show up.
- The formatting for commands' documentation and examples is more consistent. There is now one component which takes the command description, its examples (if it has some), and the related commands (if they exist), and does the rendering. There are a few exceptions with extensive documentation that haven't been touched.
- The short-notation of options has been replaced with long-notation for clarity (`-a` replaced by `--app`)


yohann-bacha:
This PR will complete #869 and #870

Soulou (changes requested):
```
└> scalingo create
NAME:
   scalingo create - Create a new app
  
USAGE:
   scalingo create [command options] app-id
  
CATEGORY:
   Global
  
DESCRIPTION:
   Create a new app

   Examples
     $ scalingo create mynewapp
     $ scalingo --remote "staging" create mynewapp
  
OPTIONS:
   --remote value     Remote to add to your current git repository (default: "scalingo")
   --buildpack value  URL to a custom buildpack that Scalingo should use to build your application
   --region value     Name of the region to use
   --help, -h         show help (default: false)
```

That's great, but

```
└> scalingo create --help
Incorrect Usage: flag: help requested

Fail to run command: flag: help requested
```

Doesn't work as expected, I would expect to see the same help, can you do something about that? (and maybe other commands)

```
└> scalingo --app nginx-12 logs --help
Incorrect Usage: flag: help requested

Fail to run command: flag: help requested
```

Thanks
```

yohann-bacha:
This is actually an opened issue that this PR doesn't solve for the moment : https://github.com/Scalingo/cli/issues/865

Soulou (approved):

PR #873, opened 2023-01-27
fix: use a current stacks in help

btrd:

No need of a changelog entry for this IMHO


yohann-bacha (approved):
I listed the possibles stacks with `scalingo stacks`, the `scalingo-20` is the only available, so for me it's ok

PR #932, opened 2023-04-04
fix(consent) Add consent for more commands

### LINKED ISSUE #931: [Consent] Add consent for more commands
All commands that edits resources will now ask for consent

johnsudaar:
Fix #931


EtienneM (commented):
Instead of checking the consent for an allow-list of commands, shouldn't we check it for all commands by default and define a list of commands in which we don't need to check for the consent?

johnsudaar:
It could be an idea but it's not that easy since in our generic command wrapper we have not yet run the current app detector nor do we know if we should. We might at term refactor to have generic app commands but i did not want to make such a refactor.

EtienneM (approved):

PR #935, opened 2023-04-13
fix: correct example in documentation

btrd:

⬆️ Do I need to update the changelog for a change this small?

Also I'm wondering if this command shouldn't use the `--addon` params like other command?

### DIFF
--- a/cmd/addons.go
+++ b/cmd/addons.go
@@ -41,7 +41,7 @@ var (
 		Category:  "Addons",
 		Flags:     []cli.Flag{&appFlag},
 		Usage:     "Provision an add-on for your application",
-		ArgsUsage: "addon-id plan",
+		ArgsUsage: "addon-name plan",
 		Description: CommandDescription{
 			Description: "Provision an add-on for your application",
 			Examples:    []string{"scalingo --app my-app addons-add mongodb mongo-starter-512"},
@@ -74,7 +74,7 @@ var (
 		ArgsUsage: "addon-id",
 		Description: CommandDescription{
 			Description: "Remove an existing addon from your app",
-			Examples:    []string{"scalingo --app my-app addons-remove mongodb"},
+			Examples:    []string{"scalingo --app my-app addons-remove addon_uuid"},
 			SeeAlso:     []string{"addons", "addons-add"},
 		}.Render(),
 
@@ -104,7 +104,7 @@ var (
 		ArgsUsage: "addon-id plan",
 		Description: CommandDescription{
 			Description: "Upgrade an addon attached to your app",
-			Examples:    []string{"scalingo --app my-app addons-upgrade mongodb mongo-starter-256"},
+			Examples:    []string{"scalingo --app my-app addons-upgrade addon_uuid mongo-starter-256"},
 			SeeAlso:     []string{"addons-plans", "addons-remove"},
 		}.Render(),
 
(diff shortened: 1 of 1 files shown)


cmd/addons.go
```diff
@@ -133,7 +133,7 @@ var (
 		ArgsUsage: "addon-id",
 		Description: CommandDescription{
 			Description: "Display information about an add-on attached to your app",
-			Examples:    []string{"scalingo --app my-app addons-info mongodb"},
+			Examples:    []string{"scalingo --app my-app addons-info addon_uuid"},
```
EtienneM, line 136: I would actually prefer if we could make the previous example work. We developed a fix for this use case for the `logs` command, we could do the same for the `addons-info`

EtienneM (commented):
> arrow_up Do I need to update the changelog for a change this small?

i would say it's OK not to have a changelog for this

yohann-bacha (changes requested):

cmd/addons.go
btrd: > We developed a fix for this use case for the logs command, we could do the same for the addons-info

`logs` is taking a `--addon` params, it's not the case here (it's what I'm proposing also).

I tried a to duplicate the `logs` behavior for the `addons-info` but didn't succesed, and I don't have the time currently to do it.

I would be in favor of accepting this PR, and wrote an issue to improve this command, in order to accept a params or a flag and a addon name or uuid.

EtienneM: OK I think you are right. I opened #937.

@yohann-bacha currently both `addon_uuid` and `addon-uuid` are in use, there is unfortunately no convention. Hence it  looks good to me.

EtienneM (approved):

cmd/addons.go
btrd: @yohann-bacha waiting for your review in order to merge this PR :)

yohann-bacha: @EtienneM My point was not about using a `_` or a `-`, this is more a problem of consistency.

Currently for every example in `addons.go`, we are giving real names for addons. As `mongodb`, and not `addon-id`, neither `addon_uuid`. 

If we use now `addon_uuid` this is fine, but we shall do this everywhere at least in the `addons.go` file.

The following point is not blocking, but I think examples should be real world commands that you can copy paste and use them. What we need as args must be in the ArgsUsage (they are already there), the examples should not be another ArgsUsage.

btrd: > As mongodb, and not addon-id, neither addon_uuid.

I think you miss-understood something, mongodb is not an id or uuid it's a addon name. Here the user need to use a value that looks like this `23w2e3rjwefsdjl;fgnlsdnfgnsdgn23` (yes I put my head on the keyboard), so I don't think we should give a  "real" example.

Could you Add a suggestion to this PR if I didn't understood what's you're proposing?

yohann-bacha: You are absolutely right, then in that case, in examples where an addon-id is required, there should always be `addon_uuid` then. Line 107, 77...

And for the addon-add command, the argusage is wrong : it is not the addon-id we are waiting for, but the addon name

btrd: > You are absolutely right, then in that case, in examples where an addon-id is required, there should always be addon_uuid then. Line 107, 77...
> And for the addon-add command, the argusage is wrong : it is not the addon-id we are waiting for, but the addon name

Indeed it seems there are other errors in this file :/

btrd: @yohann-bacha Thanks for the update !

@EtienneM can you review the new commit?

EtienneM (approved):

PR #939, opened 2023-04-30
feat: add multi architecture nix flake

cimm:
[Nix](https://nixos.org/) is a package manager, think `homebrew` on macOS or `apt` on Debian/Ubuntu. The Scalingo CLI isn't yet available in the nix repository, though there is a [pending pull request](https://github.com/NixOS/nixpkgs/pull/224858).

[Flakes](https://xeiaso.net/blog/nix-flakes-1-2022-02-21) are a newer nix feature and can run or install software directly from online sources. This pull request includes a flake for the Scalingo CLI 1.28.2 client. With only the nix package manager installed, you can run the Scalingo CLI client directly without installing anything:

```
$ nix run github:cimm/scalingo-cli/add-nix-flake#scalingo -- --version
Scalingo Client version 1.28.2
```

The command could be simplified to `nix run github:Scalingo/cli -- --version` once this pull request has been merged.

Flakes can also be used to develop and build applications without having to install development dependencies. `nix build github:cimm/scalingo-cli/add-nix-flake`, for example, builds the client in the `./results/bin` directory, even when go is not installed on the system.

Anyway, I use it daily to work on our webapps hosted at Scalingo, so it might be useful for others as well. Feel free to close the pull request if this is out of scope, no hard feelings.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,59 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1682929865,
+        "narHash": "sha256-jxVrgnf5QNjO+XoxDxUWtN2G5xyJSGZ5SWDQFxMuHxc=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "f2e9a130461950270f87630b11132323706b4d91",
+        "type": "github"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "utils": "utils"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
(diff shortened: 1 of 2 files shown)


EtienneM (commented):
Hi,

Thanks a lot for your contribution. We need to discuss internally whether we want to maintain this Nix support or not. We will get back to you as soon as possible.

josephpage:
I find this PR very interesting, with Nix gaining in popularity, but I wonder how the CLI version will be kept up to date over time. I have the impression that, in its current state, it will require manual intervention, which is not ideal. Perhaps it should be integrated into the release tools.

cimm:
You are right, there will be some added maintenance burden when this would be merged on master. Since the Scalingo CLI recently upgraded to 1.29.1 I changed the flake file reflect this, see commit 42dce19. This should give you an idea of the work involved for a typical upgrade.

nlewo:
Thanks to @cimm, this CLI is now part of the [nixpkgs repository](https://github.com/NixOS/nixpkgs/pull/224858) and a bot will submit PR on new releases. So, I don't think you would still need to maintain the build here and this PR could be closed.

If you were Nix users, the `devShell` would still be pretty useful since it provides a reproducible dev environment in a single command line. Note it could also be a first step into the nixpkgs ecosystem which is already used by [some companies](https://blog.replit.com/nix) "close" to your domain [end of Nix propaganda].

cimm:
Thanks for the merge help @nlewo! Indeed, we no longer need the flake version since the Scalingo CLI is now available in the Nix repository, I will close this pull request.

PR #948, opened 2023-05-11
fix(cli): return an exit code equal to 1 when a command or flag is malformed

### LINKED ISSUE #947: No exit code is returned when command 'integration-link-manual-deploy' execution has failed
Actually no exit code is returned when the following command fails : 
` scalingo --app my-app integration-link-manual-deploy --branch "my-branch"`
`Incorrect Usage: flag provided but not defined: -branch`

When running `echo $?` the result is 0 (successful) 

**Expected result**
When the command execution fails, the exit code shoud be other than 0.
`echo $?` should be other than 0.

ℹ️  I have seen it on this command but maybe is common to all commands.

Soulou:
Fixes #947


scalingo/main.go
```diff
@@ -127,14 +129,24 @@ func main() {
...
-	if err := app.RunContext(ctx, os.Args); err != nil {
+	err := app.RunContext(ctx, os.Args)
+	if err != nil {
 		fmt.Println("Fail to run command:", err)
 	}
+
+	// We want to display to the user if a new version is available
+	// Whatever the success of the execution of their command is.
+	updateCheckErr := update.Check()
+	if err != nil {
```
yohann-bacha, line 145: **issue:** isn't it updateCheckErr you should check?

yohann-bacha (changes requested):

scalingo/main.go
Soulou: You're 100% right

yohann-bacha (approved):

teddyanoloc:
thank you very much for this quick fix ! 👍

PR #950, opened 2023-05-20
Set env variables from a file

### LINKED ISSUE #782: [env] Have an option to set variables from a .env file on the `env-set` command
Ability to set environment variables from a .env to an application in one command

Request from a customer: https://app.intercom.com/a/apps/w4oogu7s/inbox/inbox/3379738/conversations/12375700081215

francois2metz:
This add a new flag `--file` to the `env-set` subcommand to read from an env file or read from stdin.

This should fix #782

### DIFF
--- a/env/edit.go
+++ b/env/edit.go
@@ -28,23 +28,29 @@ func Add(ctx context.Context, app string, params []string, filePath string) erro
 		return scalingoerrors.Notef(ctx, err, "read .env file")
 	}
 
-	variablesFromCmdLine, err := readFromCmdLine(ctx, params)
+	variables, err := readFromCmdLine(ctx, variablesFromFile, params)
 	if err != nil {
 		return scalingoerrors.Notef(ctx, err, "read variables from command line")
 	}
 
-	variables := mergeVariables(variablesFromFile, variablesFromCmdLine)
+	scalingoVariables := scalingo.Variables{}
+	for name, value := range variables {
+		scalingoVariables = append(scalingoVariables, &scalingo.Variable{
+			Name:  name,
+			Value: value,
+		})
+	}
 
 	c, err := config.ScalingoClient(ctx)
 	if err != nil {
 		return scalingoerrors.Notef(ctx, err, "get Scalingo client")
 	}
-	_, _, err = c.VariableMultipleSet(ctx, app, variables)
+	_, _, err = c.VariableMultipleSet(ctx, app, scalingoVariables)
 	if err != nil {
 		return scalingoerrors.Notef(ctx, err, "set multiple variables")
 	}
 
-	for _, variable := range variables {
+	for _, variable := range scalingoVariables {
 		fmt.Printf("%s has been set to '%s'.\n", variable.Name, variable.Value)
 	}
 	fmt.Println("\nRestart your containers to apply these environment changes on your application:")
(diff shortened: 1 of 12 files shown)


env/edit.go
```diff
@@ -6,11 +6,13 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"os"
 
 	"gopkg.in/errgo.v1"
 
 	"github.com/Scalingo/cli/config"
 	"github.com/Scalingo/go-scalingo/v6"
+	"github.com/joho/godotenv"
```
EtienneM, line 15: I am kind of reluctant to introduce an external dependency for this use case. What do you think about this @Soulou?

Soulou: @EtienneM Actually, it doesn't bring a lot of value to re-implement it and we can understand the need related to this PR.

The question is more: is the dependency safe enough?

* Is it maintained? (Number of releases, developers, etc.)
* Is CI configured?
* Quick audit of the code base

To me these signals are green-ish, and would go for it.

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/google/go-github/v47 v47.1.0
 	github.com/gorilla/websocket v1.5.0
 	github.com/gosuri/uilive v0.0.4
+	github.com/joho/godotenv v1.5.1
```
EtienneM, line 20: Dependency must be vendored

env/edit.go
EtienneM: I'm not sure I agree with these arguments but you are the boss :p

This dependency seems well maintained with good programming practice, let's go :rocket:

env/edit.go
```diff
@@ -20,18 +22,17 @@ var (
...
-			Value: value,
-		})
+	err := readFromFile(&variables, filePath)
+	if err != nil {
+		return err
+	}
+
+	err = readFromCmdLine(&variables, params)
+	if err != nil {
+		return err
```
EtienneM, line 35: best practice: I would prefer having both `readFromFile` and `readFromCmdLine` returning a `scalingo.Variables` then merging them after these calls. We are not keen on modifying functions input arguments at Scalingo.

best practice: please wrap the returned error with `github.com/Scalingo/go-utils/errors/v2.Notef`

question: what if `.env` contains a variable which is also defined on the command line? Is it defined twice in `variables`? Shouldn't we make sure it's only once in `variables`?

env/edit.go
```diff
@@ -106,3 +107,43 @@ func parseVariable(param string) (string, string) {
...
+		*variables = append(*variables, &scalingo.Variable{
+			Name:  name,
+			Value: value,
+		})
+	}
+	return nil
+}
+
+func readFromFile(variables *scalingo.Variables, filePath string) error {
+	if len(filePath) > 0 {
```
EtienneM, line 127: best practice: we prefer exiting the function as early as possible.
```suggestion
	if len(filePath) == 0 {
		return nil
	}
```

go.mod
francois2metz: I forgot that, done!

env/edit.go
francois2metz: All done!

I fixed the case when a variable has been defined twice. The command line variable will take precedence from the env file.

PR #954, opened 2023-05-26
feat(install.sh): change default target directory for Apple Silicon

### LINKED ISSUE #611: Installer should use a different default directory on macOS + arm64 (Apple Silicon)
With an Intel processor based Mac, Homebrew is installed in `/usr/local` and this is also the default installation directory for the Scalingo CLI binary.

With an Apple processor (arch = arm64) Homebrew is installed in `/opt/homebrew` which makes Scalingo CLI install script crash.

However I can confirm that running the `amd64` binary described in https://github.com/Scalingo/cli/issues/609 works perfectly (via Rosetta 2, automatically) 

As `uname -m` returns `arm64` - see https:/stackoverflow.comquestions65259300detect-apple-silicon-from-command-line about this - maybe you could integrate this, and change the default instillation directory accordingly ?

Currently the default directory it is hard-coded:

```shell
target_dir="${target_dir:-/usr/local/bin}""
```

https://github.com/Scalingo/cli/blob/9a45007ad95ce8b463da1f39947d6ac765973d55/dists/install.sh

Perhaps we could evolve to this pseudo code :

```shell
default_target_dir="/usr/local/bin"
box $arch in
  arm64)
    default_target_dir="/opt/homebrew/bin"
    ;;
esac

target_dir="${target_dir:-$default_target_dir}""
```

Otherwise you could detect if `/opt/home/brew` exists and use it for the default installation directory.
If not you could fall back to the default `/usr/local` ?

SCKevinO:
Fix #611 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * feat(install.sh): handling word answers as user inputs for install script
 * fix(region_migrations): help message for `migration-abort` ([PR#951](https://github.com/Scalingo/cli/pull/951))
 * feat(env): set variables from an env file ([PR#950](https://github.com/Scalingo/cli/pull/950))
+* feat(install.sh): change default target directory to /opt/homebrew/bin for Apple Silicon
 
 ### 1.28.2
 
--- a/dists/install.sh
+++ b/dists/install.sh
@@ -45,7 +45,7 @@ main() {
     echo "Options:"
     echo "  -h, --help             displays help and exits"
     echo "  -i, --install-dir DIR  Scalingo client installation directory, creating it if"
-    echo "                         necessary (defaults to /usr/local/bin)"
+    echo "                         necessary (defaults to /usr/local/bin or /opt/homebrew/bin for Apple Silicon)"
     echo "  -y, --yes              overwrites previously installed Scalingo client"
     echo
   }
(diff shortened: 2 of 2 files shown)


EtienneM (approved):
Thanks :)

josephpage:
Hello, sorry I'm only reading the PR now, but I'm wondering about using a folder that only exists when Homebrew is installed for an install script that's going to be used by those who don't usually use Homebrew and on whom it probably won't be installed. Or maybe I'm missing something and I'm sorry to bother you.

josephpage:
Thanks for the reply, I understand your reasoning better.
Installing a binary in a directory managed by Homebrew could lead to an error during a future installation with Homebrew. That's my only concern.

In fact, Homebrew is so popular on the Mac that most tools now only offer installation with Homebrew or with an nerdy alternative method for those who want full control : a binary to download and move yourself. For example:
- Github CLI: https://github.com/cli/cli#installation
- kubectl : https://kubernetes.io/docs/tasks/tools/install-kubectl-macos/
- Heroku : https://devcenter.heroku.com/articles/heroku-cli#install-with-an-installer
- Scaleway CLI : https://github.com/scaleway/scaleway-cli#homebrew
- Terraform : https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli#install-terraform
- jq : https://jqlang.github.io/jq/download/
And the list goes on...

The cleanest solution might even be to deprecate this install script for Macs in favor of the Homebrew formula.

PR #974, opened 2023-06-16
fix(term): use os.Stat to get TTY information

### LINKED ISSUE #857: Replace `github.com/andrew-d/go-termutil` with a maintained lib
Or maybe even the standard library

curzolapierre:

Fix #857


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### To be Released
 
+* chore(term): remove go-termutil, use `Stat` instead.
```
EtienneM, line 5: nitpick:

```suggestion
* chore(term): remove `github.com/andrew-d/go-termutil`, use standard library instead
```

term/term_unix.go
```diff
@@ -21,7 +20,9 @@ func IsTerminal(f *os.File) bool {
 }
 
 func IsATTY(f *os.File) bool {
-	return termutil.Isatty(f.Fd())
+	stat, _ := f.Stat()
```
EtienneM, line 23: I don't think it's OK to ignore the error here. Maybe return false in such cases? I have the feeling that it's what is currently done in the dependency: 

https://github.com/andrew-d/go-termutil/blob/009166a695a2f516c749a26b4ac1f183d89aa336/isatty_nix.go#L19

EtienneM (commented):
A couple of questions, otherwise if it works LGTM :)

EtienneM (approved):

PR #988, opened 2023-07-31
feat(install.sh): verify the archive checksum

### LINKED ISSUE #422: Release files should have a checksum
And this checksum should be checked by the install script.

EtienneM:
If valid, output is:

```
-----> Downloading Scalingo client...  DONE
-----> Verifying the checksum...  VALID
[...]
```

If invalid, output is:

```
-----> Downloading Scalingo client...  DONE
-----> Verifying the checksum...  INVALID
 !     Checksums don't match ('3fdec56b8647953744fe4c9094dd7b0a0952b1cb9416bc7e315cde00b8db557a' != '3fdec56b8647953744fe4c9094dd7b0a0952b1cb9416bc7e315cde00b8db557a')
 !     You may want to retry to install the Scalingo CLI. If the problem persists, please contact our support team.
```

I'm not sure we want to keep displaying the checksums in the error message, what do you think?

Fix #422

### DIFF
--- a/dists/install.sh
+++ b/dists/install.sh
@@ -10,7 +10,7 @@ main() {
   }
 
   error() {
-    echo -en " !     $*"
+    echo -en " !     $*" >&2
   }
 
   warn() {
@@ -112,8 +112,8 @@ main() {
 
   version=$(curl --silent https://cli-dl.scalingo.com/version | tr -d ' \t\n')
   if [ -z "$version" ]; then
-    echo "-----> Fail to get the version of the CLI" >&2
-    echo "You probably have an old version of curl. Please check your curl version and update accordingly." >&2
+    error "Fail to get the version of the CLI\n"
+    error "You probably have an old version of curl. Please check your curl version and update accordingly.\n"
     exit 1
   fi
 
@@ -124,8 +124,8 @@ main() {
   status "Downloading Scalingo client...  "
   curl --silent --fail --location --output ${tmpdir}/${archive_name} ${url}
   if [ ! -f ${tmpdir}/${archive_name} ]; then
-    echo "" >&2
-    echo "-----> Fail to download the CLI archive" >&2
+    echo ""
+    error "Fail to download the CLI archive\n"
     exit 1
   fi
   echo "DONE"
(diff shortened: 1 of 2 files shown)


dists/install.sh
```diff
@@ -112,30 +112,44 @@ main() {
 
   version=$(curl --silent https://cli-dl.scalingo.com/version | tr -d ' \t\n')
   if [ -z "$version" ]; then
-    echo "-----> Fail to get the version of the CLI" >&2
-    echo "You probably have an old version of curl. Please check your curl version and update accordingly." >&2
+    error "Fail to get the version of the CLI\n"
```
EtienneM, line 115: I also updated a few lines to use the `error` function when we want to display error messages.

dists/install.sh
```diff
@@ -112,30 +112,43 @@ main() {
...
   fi
   echo "DONE"
+
+  status "Verifying the checksum...  "
+  checksums_url="https://github.com/Scalingo/cli/releases/download/${version}/checksums.txt"
+  checksum_computed=$(sha256sum ${tmpdir}/${archive_name} | cut -d " " -f1)
+  checksum_expected=$(wget -q --output-document - $checksums_url | grep $archive_name | cut -d " " -f 1)
+  if [[ "$checksum_computed" != "$checksum_expected" ]]; then
+    echo "INVALID"
+    error "Checksums don't match ('$checksum_computed' != '$checksum_expected').\n"
```
ipfaze, line 139: suggestion(non-blocking): As you said in the PR description, I'm not sure either to keep displaying the checksums.
Could it be possible to only display "Checksums don't match" but when adding `DEBUG=1`, then display the checksums?

dists/install.sh
```diff
@@ -112,30 +112,43 @@ main() {
...
+    echo ""
+    error "Fail to download the CLI archive\n"
     exit 1
   fi
   echo "DONE"
+
+  status "Verifying the checksum...  "
+  checksums_url="https://github.com/Scalingo/cli/releases/download/${version}/checksums.txt"
+  checksum_computed=$(sha256sum ${tmpdir}/${archive_name} | cut -d " " -f1)
+  checksum_expected=$(wget -q --output-document - $checksums_url | grep $archive_name | cut -d " " -f 1)
```
ipfaze, line 136: nitpick(non-blocking): Can you use the full arguments rather than the shortcut ones?

ipfaze (approved):
Some suggestion and nitpick otherwise LGTM :+1:

dists/install.sh
EtienneM: This should be a blocking comment, you are right ;)

dists/install.sh
EtienneM: There is already a `DEBUG=true` which defines `set -x`, which would display the variables content. Hence I removed my line.

PR #1007, opened 2023-09-18
feat(region): more debug logs

### LINKED ISSUE #866: Add more debug logs on region cache system
On the file `config/region.go` there not to much logs on `EnsureRegionsCache` and `GetRegion` functions. We should add more debug logs because we have probably some edge-case not handled correctly.

Related to the following level3 support: https://scalingo.slack.com/archives/C03T384J3/p1673262946393849

EtienneM:
Fix #866

### DIFF
--- a/config/region.go
+++ b/config/region.go
@@ -49,6 +49,7 @@ type GetRegionOpts struct {
 }
 
 func EnsureRegionsCache(ctx context.Context, c Config, opts GetRegionOpts) (RegionsCache, error) {
+	debug.Println("[Regions] Ensure cache is filled")
 	var regionsCache RegionsCache
 	fd, err := os.Open(c.RegionsCachePath)
 	if err == nil {
@@ -63,6 +64,7 @@ func EnsureRegionsCache(ctx context.Context, c Config, opts GetRegionOpts) (Regi
 
 	var client *scalingo.Client
 	if opts.SkipAuth {
+		debug.Println("[Regions] Create an unauthenticated client to the authentication service")
 		client, err = ScalingoUnauthenticatedAuthClient(ctx)
 		if err != nil {
 			return RegionsCache{}, errgo.Notef(err, "fail to create an unauthenticated client")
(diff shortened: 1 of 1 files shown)


config/region.go
```diff
@@ -49,9 +49,12 @@ type GetRegionOpts struct {
 }
 
 func EnsureRegionsCache(ctx context.Context, c Config, opts GetRegionOpts) (RegionsCache, error) {
+	debug.Println("[Regions] Ensure cache is filled")
 	var regionsCache RegionsCache
 	fd, err := os.Open(c.RegionsCachePath)
-	if err == nil {
+	if err != nil {
+		debug.Printf("[Regions] Fail to open the cache: %v\n", err)
+	} else {
```
EtienneM, line 57: We are more used to check `err != nil` in Go. So I'm refactoring to use this pattern here.

config/region.go
```diff
@@ -93,18 +98,20 @@ func EnsureRegionsCache(ctx context.Context, c Config, opts GetRegionOpts) (Regi
...
 	fd, err = os.OpenFile(c.RegionsCachePath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0750)
-	if err == nil {
-		json.NewEncoder(fd).Encode(regionsCache)
-		fd.Close()
-	} else {
-		debug.Printf("[Regions] Failed to save the regions cache: %v\n", err)
+	if err != nil {
+		debug.Printf("[Regions] Failed to save the cache: %v\n", err)
+		return regionsCache, nil
 	}
```
EtienneM, line 114: We are more used to check `err != nil` in Go. So I'm refactoring to use this pattern here.

config/region.go
ipfaze: Question: Why don't we use `defer fd.Close()`?
It seems more secure IMO.

ipfaze: Question: Why don't we use defer fd.Close()?
It seems more secure IMO.

ipfaze (changes requested):

config/region.go
EtienneM: I don't know. Maybe it's useless in this case to wait longer before closing the file descriptor: we decode its content then immediately close it.

brandon-welsch (approved):
LGTM

config/region.go
ipfaze: LGTM Then :ok_hand:

ipfaze (approved):
:rocket:

PR #1009, opened 2023-09-18
refactor(auth): replace use of deprecated `ErrOTPRequired`

EtienneM:
Related to https://github.com/Scalingo/go-scalingo/issues/360

Fix https://github.com/Scalingo/go-scalingo/issues/324
Fix https://github.com/Scalingo/go-scalingo/issues/326

### DIFF
--- a/config/auth.go
+++ b/config/auth.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	errgo "gopkg.in/errgo.v1"
 
 	"github.com/Scalingo/cli/config/auth"
 	appio "github.com/Scalingo/cli/io"
@@ -28,7 +27,7 @@ var (
 type CliAuthenticator struct{}
 
 var (
-	ErrUnauthenticated = errgo.New("user unauthenticated")
+	ErrUnauthenticated = errors.New("user unauthenticated")
 )
 
 func Auth(ctx context.Context) (*scalingo.User, string, error) {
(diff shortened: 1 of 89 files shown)


config/auth.go
```diff
@@ -11,12 +11,12 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	errgo "gopkg.in/errgo.v1"
```
EtienneM, line 14: I removed this dependency from this file and it had a lot of side effects, sorry.. :sweat_smile:

config/auth.go
```diff
@@ -43,7 +43,7 @@ func Auth(ctx context.Context) (*scalingo.User, string, error) {
 			if err == nil {
 				break
 			} else if scalingoerrors.RootCause(err) == io.EOF {
```
ipfaze, line 45: issue(non-blocking): Can you change the deprecated function `.RootCause()` to the function `.Is()`

config/auth.go
```diff
@@ -11,12 +11,12 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
```
ipfaze, line 13: suggestion: Is it possible to replace all existing `errors` to `scalingoerrors`? 
It will freed us from the use of another deprecated package

ipfaze (changes requested):
Still some improvements possible but otherwise LGTM

config/auth.go
EtienneM: It's not possible. We declare some public variable, for instance on line 24:

```go
var (
	ErrAuthenticationFailed = errors.New("authentication failed")
)
```

`scalingoerrors` requires a context and we have no context available here.

ipfaze (approved):
LGTM then :+1:

PR #1011, opened 2023-09-26
fix errors -> error

adrienpoly:
I was getting this error when installing Scalingo CLI 

Capture d’écran 2023-09-26 à 09 45 50


EtienneM (approved):
Thanks a lot :+1:

Issue #1012, opened 2023-09-26
[MacOS] Install script failed due to usage of sha256sum


curzolapierre:
Following the addition of checksum (https://github.com/Scalingo/cli/pull/988) check, the installation of the script is not working anymore on MacOS lacking some tools (i.e. `sha256sum`, `wget` not installed by default).
 
A test has been added on https://github.com/Scalingo/cli/pull/1002 (issue: #1001), but it is probably not what we want to.
Either the script installs the tools or it uses other tools.

E.g. replace `sha256sum` by a function that use openssl:
```bash
function sha256sum() { openssl sha256 "$@" | awk '{print $2}'; }
```
e.g. replace `wget` by `curl`

PR #1013, opened 2023-09-26
fix(install): remove wget and sha256sum dependencies

curzolapierre:
Fixes https://github.com/Scalingo/cli/issues/1012

Tested on Benjamin's Mac

### DIFF
--- a/dists/install.sh
+++ b/dists/install.sh
@@ -24,9 +24,9 @@ main() {
   clean_install() {
     tmpdir=$1
 
-    rm -r $tmpdir
+    rm -r "$tmpdir"
     # If installed through one line install, remove script
-    if [ "x$0" = "xinstall" ] ; then
+    if [ "$0" = "install" ] ; then
       rm "$0"
     fi
   }
@@ -35,7 +35,7 @@ main() {
     info "$* [Y/n] "
     while true; do
         read answer
-        case $(echo "$answer" | tr "[A-Z]" "[a-z]") in
+        case $(echo "$answer" | tr '[:upper:]' '[:lower:]') in
         y|yes|"" ) return 0;;
         n|no     ) return 1;;
         esac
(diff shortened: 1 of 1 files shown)


dists/install.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+function sha256sum() {
+  shasum -a 256 "$@"
+}
+
```
curzolapierre, line 6: This is the actual fix

dists/install.sh
```diff
@@ -131,22 +127,19 @@ main() {
...
-  check_command_exists sha256sum
-  check_command_exists wget
-
   status "Verifying the checksum...  "
   checksums_url="https://github.com/Scalingo/cli/releases/download/${version}/checksums.txt"
   # Use cut's short parameter to be compatible with MacOS: https://ss64.com/osx/cut.html
-  checksum_computed=$(sha256sum ${tmpdir}/${archive_name} | cut -d" " -f1)
-  checksum_expected=$(wget --quiet --output-document - $checksums_url | grep $archive_name | cut -d" " -f1)
+  checksum_computed=$(sha256sum "${tmpdir}/${archive_name}" | cut -d" " -f1)
+  checksum_expected=$(curl --silent --location "$checksums_url" | grep "$archive_name" | cut -d" " -f1)
```
curzolapierre, line 142: This is the actual fix

dists/install.sh
EtienneM: question: from what I understood of the actual problem, the call to `sha256sum` works as this script used to work on Benjamin's Mac. It's the introduction of the function `check_command_exists` that made the install fail. Isn't it?

dists/install.sh
```diff
@@ -58,14 +54,14 @@ main() {
     echo
   }
 
-  if [ "x$DEBUG" = "xtrue" ] ; then
+  if [ -n "$DEBUG" ] ; then
```
EtienneM, line 57: it's not equivalent. Why don't we keep the equivalent?
```suggestion
  if [ "$DEBUG" = "true" ] ; then
```

dists/install.sh
EtienneM: question: I have the same question as the one raised above. From what I understand of the issue, the problem is not the usage of `wget` that is here for years according to the Git history. The problem is the call to `check_command_exists`.

dists/install.sh
curzolapierre: Not really, it never worked on Benjamin's Mac since the addition of the call to `sha256sum`.
This binary seems not installed by default on MacOS.

dists/install.sh
curzolapierre: Yes `check_command_exists` blocked user that didn't have `wget` installed.
But in the same way as this command failed due to missing binary.

I found it more consistent to use curl as we are already asking to use for the download https://doc.scalingo.com/platform/cli/start

dists/install.sh
curzolapierre: It's not equivalent, I change it to be more consistent with the usage of the `DEBUG` of the CLI itself (which can be used with `DEBUG=1`).

EtienneM (approved):
Thanks for your answers :)

PR #1015, opened 2023-10-11
fix(install): if shasum not available fall back on sha256sum

curzolapierre:
~~- [ ] Add a changelog entry in the section "To Be Released" of CHANGELOG.md~~


yohann-bacha (approved):
If it works, LGTM!

PR #1019, opened 2023-11-13
feat(cmd): add db's user list

### LINKED ISSUE #1018: [MAINTENANCE-WINDOWS | STORY-150] Database user management
STORY-150

```bash
# List
$ scalingo --app my-app --addon my-addon-uuid database-list-users
# Delete
$ scalingo --app my-app --addon my-addon-uuid database-delete-user metabase-user
# Create
$ scalingo --app my-app --addon my-addon-uuid database-create-user my-user
```

curzolapierre:
- [x] Update go-scalingo once it has been released (https://github.com/Scalingo/go-scalingo/pull/365)

## List

```sh
$ scalingo database-list-users --help
NAME:
   scalingo database-list-users - Print database's users

USAGE:
   scalingo database-list-users command [command options] [arguments...]

DESCRIPTION:
   List the users of a database

   Only available on [postgresql elasticsearch influxdb]

   Example
     $ scalingo --app myapp --addon addon-uuid database-list-users

OPTIONS:
   --app value, -a value  Name of the current app (default: "<name>")
   --addon value          ID of the current addon (default: "<addon_id>")
   --region value         Name of the region to use
   --help, -h             show help
```

## Delete

```sh
$ scalingo database-delete-user --help
NAME:
   scalingo database-delete-user - Delete a database's user

USAGE:
   scalingo database-delete-user command [command options] user

DESCRIPTION:
   Delete the given user of a database

   Only available on [postgresql elasticsearch influxdb]

   Example
     $ scalingo --app myapp --addon addon-uuid database-delete-user my_user

OPTIONS:
   --app value, -a value  Name of the current app (default: "<name>")
   --addon value          ID of the current addon (default: "<addon_id>")
   --region value         Name of the region to use
   --help, -h             show help
```

## Create

```sh
$ scalingo database-create-user --help
NAME:
   scalingo database-create-user - Create new database user

USAGE:
   scalingo database-create-user command [command options] user

DESCRIPTION:
   Create new database user

   Only available on [postgresql elasticsearch influxdb]

   Examples
     $ scalingo --app myapp --addon addon-uuid database-create-user my_user
     $ scalingo --app myapp --addon addon-uuid database-create-user my_user --read-only

OPTIONS:
   --app value, -a value  Name of the current app (default: "<name>")
   --addon value          ID of the current addon (default: "<addon_id>")
   --read-only            Create a user with read-only rights (default: false)
   --region value         Name of the region to use
   --help, -h             show help
```

Fixes #1018


cmd/databases.go
```diff
@@ -138,6 +140,108 @@ var (
...
+		Flags:     []cli.Flag{&appFlag, &addonFlag},
+		Description: CommandDescription{
+			Description: "Delete the given user of a database",
+			Examples: []string{
+				"scalingo --app myapp --addon addon-uuid database-delete-user my_user",
+			},
+		}.Render(),
+
+		Action: func(c *cli.Context) error {
+			if c.Args().Len() != 1 {
```
yohann-bacha, line 183: **issue:** in this case, an error should be returned. the exit code 0 means the operation was done successfully, which is not the case here

cmd/databases.go
```diff
@@ -138,6 +140,108 @@ var (
...
+		Description: CommandDescription{
+			Description: "Create new database user",
+			Examples: []string{
+				"scalingo --app myapp --addon addon-uuid database-create-user my_user",
+				"scalingo --app myapp --addon addon-uuid database-create-user my_user --read-only",
+			},
+		}.Render(),
+
+		Action: func(c *cli.Context) error {
+			if c.Args().Len() != 1 {
```
yohann-bacha, line 224: **issue:** same issue as before, this should return an error

db/users/utils.go
```diff
@@ -0,0 +1,33 @@
...
+	"errors"
+	"strings"
+
+	"github.com/Scalingo/cli/config"
+	scErrors "github.com/Scalingo/go-utils/errors/v2"
+)
+
+var ErrDatabaseNotSupportUserManagement = errors.New("Error: DBMS does not support user management")
+
+func isDatabaseHandlesUserManagement(ctx context.Context, app, addonUUID string) (bool, error) {
```
yohann-bacha, line 14: **nitpick:** that's a weird english

```suggestion
func doesDatabaseHandleUserManagement(ctx context.Context, app, addonUUID string) (bool, error) {
```

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+		fmt.Printf("User \"%s\" created with password \"%s\".\n", databaseUsers.Name, password)
+		return nil
+	}
+
+	fmt.Printf("User \"%s\" created.\n", databaseUsers.Name)
+	return nil
+}
+
+func askForPassword(ctx context.Context) (string, string, error) {
+	fmt.Printf("Password: (Will be generated if left empty) ")
```
yohann-bacha, line 78: **nitpick:**
```suggestion
	fmt.Printf("Password (will be generated if left empty): ")
```

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return "", "", scErrors.Wrap(ctx, err, "read password")
+	}
+
+	fmt.Printf("\nPassword Confirmation: ")
+	byteConfirmedPassword, err := term.ReadPassword(int(os.Stdin.Fd()))
+	if err != nil {
+		return "", "", scErrors.Wrap(ctx, err, "read password confirmation")
+	}
+	fmt.Printf("\n")
```
yohann-bacha, line 90: **nitpick:** 
```suggestion
	fmt.Println("")
```

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+
+	return string(bytePassword), string(byteConfirmedPassword), nil
+}
+
+func isPasswordValid(password, confirmedPassword string) (string, bool) {
+	if password != confirmedPassword {
+		return "Password confirmation don't watch", false
```
yohann-bacha, line 97: **issue:** 
```suggestion
		return "Password confirmation doesn't match", false
```

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if len(password) < 8 || len(password) > 64 {
+		return "Password must be between 8 and 64 characters", false
```
yohann-bacha, line 100: **suggestion (non-blocking):**
```suggestion
		return "Password must contain between 8 and 64 characters", false
```

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return "", true
+}
+
+func isUsernameValid(username string) (string, bool) {
+	if len(username) < 6 || len(username) > 32 {
+		return "name must be between 6 and 32 characters", false
```
yohann-bacha, line 107: **suggestion (non-blocking):**
```suggestion
		return "name must contain between 6 and 32 characters", false
```

db/users/utils.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+	"errors"
+	"strings"
+
+	"github.com/Scalingo/cli/config"
+	scErrors "github.com/Scalingo/go-utils/errors/v2"
+)
+
+var ErrDatabaseNotSupportUserManagement = errors.New("Error: DBMS does not support user management")
```
yohann-bacha, line 12: **issue:** not a fan of making a constant of an error, as we never do this anywhere in our code. we often have a same base error such as "user cannot be fetched", but we do not factorize them

db/users/utils.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return false, scErrors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	addon, err := addonsClient.AddonShow(ctx, app, addonUUID)
+	if err != nil {
+		return false, scErrors.Wrap(ctx, err, "get the addon to check user management support")
+	}
+
+	for _, supportedAddon := range supportedAddons {
```
yohann-bacha, line 26: **praise:** i like the simple use of plurals and singular in a foreach

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+		return scErrors.Wrap(ctx, err, "ask for password")
+	}
+
+	passwordValidation, ok := isPasswordValid(password, confirmedPassword)
+	if !ok && password != "" {
+		io.Error(passwordValidation)
+		return nil
+	}
+
+	isPasswordGenerated := false
```
yohann-bacha, line 44: **issue (blocking?):** the password generation logic is made backend sided. do we want to handle it on the client side?

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	fmt.Printf("User \"%s\" created.\n", databaseUsers.Name)
+	return nil
+}
+
+func askForPassword(ctx context.Context) (string, string, error) {
+	fmt.Printf("Password: (Will be generated if left empty) ")
+
+	bytePassword, err := term.ReadPassword(int(os.Stdin.Fd()))
```
yohann-bacha, line 80: **issue:** no need to specify the type of the variable via its name
```suggestion
	password, err := term.ReadPassword(int(os.Stdin.Fd()))
```

db/users/create.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return "", "", scErrors.Wrap(ctx, err, "read password")
+	}
+
+	fmt.Printf("\nPassword Confirmation: ")
+	byteConfirmedPassword, err := term.ReadPassword(int(os.Stdin.Fd()))
```
yohann-bacha, line 86: **issue:** same, no need for specifying the type in the variable name

yohann-bacha (changes requested):
nice job! some small issues, but nothing too serious

cmd/databases.go
yohann-bacha: wtf you're in fact right, but that is weird.. i really thought that there were a client that pointed this behaviour, and that we should handle it differently. this should return an exit code 1, but you are right : consistency is more important

db/users/utils.go
yohann-bacha: > just in this repo I can see some

not really sure about that one as I never saw one in my peregrinations in all our go code. in fact, i like the idea that we centralize our errors, and then identify them more easily. though, i see this more as a consistency issue. if we do this here, we should do it every where errors are duplicated. 

this is, though, not blocking

db/users/create.go
yohann-bacha: well that could have remove some complexity client sided, but this isn't really an issue and is not blocking, do as your heart wishes

yohann-bacha (approved):
well the last nitpicks are not enough to block my approval, so LGTM! nice job

PR #1044, opened 2024-02-19
fix(autocomplete): do not show update message on autocomplete

### LINKED ISSUE #1043: [Deprecation Warning] Do not show deprecation warning on auto completion API
The deprecation warning on the auto completion API makes it completely unusable.

Examples:
image
image

curzolapierre:

Fixes #1043

### DIFF
--- a/scalingo/main.go
+++ b/scalingo/main.go
@@ -139,11 +139,16 @@ func main() {
 		fmt.Println("Fail to run command:", err)
 	}
 
-	// We want to display to the user if a new version is available
-	// Whatever the success of the execution of their command is.
-	updateCheckErr := update.Check()
-	if updateCheckErr != nil {
-		debug.Println("Failed to check if executable should be updated", updateCheckErr)
+	// Do not show update check during autocomplete
+	if bashComplete {
+		debug.Println("Do not check available update during autocompletion")
+	} else {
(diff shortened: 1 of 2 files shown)


scalingo/main.go
```diff
@@ -139,11 +139,14 @@ func main() {
...
-	if updateCheckErr != nil {
-		debug.Println("Failed to check if executable should be updated", updateCheckErr)
+	// Do not show update check during autocomplete
+	if !bashComplete {
+		// We want to display to the user if a new version is available
+		// Whatever the success of the execution of their command is.
+		updateCheckErr := update.Check()
+		if updateCheckErr != nil {
+			debug.Println("Failed to check if executable should be updated", updateCheckErr)
+		}
```
EtienneM, line 149: suggestion(non-blocking): more log so that we understand what happens
issue: does it really display the `updateCheckErr` error?

```suggestion
	if bashComplete {
		debug.Println("Do not check available update during autocompletion")
	} else {
		// We want to display to the user if a new version is available
		// Whatever the success of the execution of their command is.
		updateCheckErr := update.Check()
		if updateCheckErr != nil {
			debug.Printf("Failed to check if executable should be updated: %v\n", updateCheckErr)
		}
```

EtienneM (approved):

PR #1045, opened 2024-02-23
fix(users): user feedback messages improvements

curzolapierre:
~~- [ ] Add a changelog entry in the section "To Be Released" of CHANGELOG.md~~

### DIFF
--- a/db/users/update.go
+++ b/db/users/update.go
@@ -31,21 +31,24 @@ func UpdateUser(ctx context.Context, app, addonUUID, username string) error {
 	if err != nil {
 		return errors.Wrap(ctx, err, "get Scalingo client")
 	}
+
 	l, err := c.DatabaseListUsers(ctx, app, addonUUID)
 	if err != nil {
 		return errors.Wrap(ctx, err, "list the database's users")
 	}
-
-	// Check if the user exists
-	var userExists bool
-	for _, user := range l {
-		if user.Name == username {
-			userExists = true
+	// Check if the userParam exists and is not protected
+	var user *scalingo.DatabaseUser
+	for _, u := range l {
+		if u.Name == username {
+			user = &u
 			break
 		}
 	}
-	if !userExists {
+	if user == nil {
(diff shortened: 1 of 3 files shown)


db/users/update.go
```diff
@@ -47,6 +47,17 @@ func UpdateUser(ctx context.Context, app, addonUUID, username string) error {
 	if !userExists {
 		return errors.New(ctx, fmt.Sprintf("User \"%s\" does not exist", username))
 	}
+	// Check if the user is protected
```
ipfaze, line 50: issue(blocking): you can avoid a loop here by saving the user in the last for loop (https://github.com/Scalingo/cli/blob/2b8d41467bf986a62428527f4232f6cd5fa4c36e/db/users/update.go#L41).
And then do you check after that.

ipfaze (changes requested):

db/users/update.go
```diff
@@ -36,18 +36,27 @@ func UpdateUser(ctx context.Context, app, addonUUID, username string) error {
...
 		}
 	}
 	if !userExists {
 		return errors.New(ctx, fmt.Sprintf("User \"%s\" does not exist", username))
 	}
 
+	if userProtected {
+		return errors.New(ctx, fmt.Sprintf("User \"%s\" is protected and cannot be updated", username))
+	}
+
```
ipfaze, line 59: For better readability I suggest:
```suggestion
    var user
    for _, u := range l {
	    if u.Name == username {
		    user = u
		    break
	    }
    }
    if user == nil {
	    return errors.New(ctx, fmt.Sprintf("User \"%s\" does not exist", username))
    }
    if user.Protected {
	    return errors.New(ctx, fmt.Sprintf("User \"%s\" is protected", username))
    }
```
WDYT?

curzolapierre: I pushed a suggestion, WDYT?

ipfaze (approved):
LGTM :+1:

PR #1051, opened 2024-03-14
fix(actions): test conditional needs

curzolapierre:

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -72,7 +72,13 @@ jobs:
   releases:
     needs: [linter-master, linter-pull-request, tests]
     name: GoReleaser Build on All OS but Windows
-    if: startsWith(github.ref, 'refs/tags/')
+    # Usage of needs implies that the job will only run if all the jobs it depends on are successful.
+    # But in our case, either linter-master or linter-pull-request will be skipped.
+    # So we need to add a condition to check if the linter-master or linter-pull-request job is successful.
+    # To do so, we use the `if` condition but we also need to add `always()` to make sure the job is run
+    # related to https://github.com/actions/runner/issues/2205
+    # since a success() is added by default and skipped jobs make success to fail we need this workaround
+    if: always() && startsWith(github.ref, 'refs/tags/') && (needs.linter-master.result == 'success' || needs.linter-pull-request.result == 'success')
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -72,7 +72,9 @@ jobs:
   releases:
     needs: [linter-master, linter-pull-request, tests]
     name: GoReleaser Build on All OS but Windows
-    if: startsWith(github.ref, 'refs/tags/')
+    # related to https://github.com/actions/runner/issues/2205
+    # since a success() is added by default and skipped jobs make success to fail we need this workaround
+    if: ${{ always() && startsWith(github.ref, 'refs/tags/') && (needs.linter-master.result == 'success' || needs.linter-pull-request.result =='success') }}
```
EtienneM, line 77: question: do we really need the surrounding `${{ .. }}`? From https://github.com/actions/runner/issues/2205#issuecomment-1534430719 it does not seem so
nitpick: you lack a space in both `if`

```suggestion
    if: ${{ always() && startsWith(github.ref, 'refs/tags/') && (needs.linter-master.result == 'success' || needs.linter-pull-request.result == 'success') }}
```

.github/workflows/publish.yml
```diff
@@ -72,7 +72,9 @@ jobs:
   releases:
     needs: [linter-master, linter-pull-request, tests]
     name: GoReleaser Build on All OS but Windows
-    if: startsWith(github.ref, 'refs/tags/')
+    # related to https://github.com/actions/runner/issues/2205
+    # since a success() is added by default and skipped jobs make success to fail we need this workaround
```
EtienneM, line 76: suggestion: you explain what the workaround is, but not what the problem is. Hence it's kind of complicated to understand the problem in the first place. Could you explain it?

.github/workflows/publish.yml
curzolapierre: The `${{ .. }}` is needed in case of Yaml reserved character but indeed here it's not the case.

.github/workflows/publish.yml
EtienneM: praise: thanks, nice explanation!

EtienneM (approved):

PR #1060, opened 2024-04-26
fix(one-off): remove async one-off 

SCedricThomas:
Related to https://github.com/Scalingo/app-scheduler/issues/886

This PR aims to remove the notion of async containers from the CLI as they are by default async  


SCedricThomas:
I don't know how I can test this in dev.  
Would you have any idea @ipfaze?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### To be Released
 
+* fix(one-off): remove async field from the run command ([PR#1060](https://github.com/Scalingo/cli/pull/1060))
```
ipfaze, line 5: issue(blocking): I'm not sure we removed the async one-off feature here. But we did remove the `async` params as it is the default behavior.

ipfaze (changes requested):
issue(blocking): I'm not sure we removed the async one-off feature here. But we did remove the `async` params as it is the default behavior.

CHANGELOG.md
SCedricThomas: Yeah :/

SCedricThomas: Is it not clear enough?

ipfaze: My bad, I missed the `field` word 🤦
It's okay!

ipfaze (approved):
LGTM 👍

PR #1071, opened 2024-07-04
[STORY-642] new api endpoint to handle password reset

sc-zenokerr:

Database user password update uses new API endpoint for generating password.

This is dependent on the merging / release of the equivalent branch in go-scalingo library.

[Story 642](https://www.notion.so/scalingooriginal/New-API-endpoint-to-handle-password-reset-662bd5bb9ec045308e5348f499055e65)

### DIFF
--- a/db/users/update_password.go
+++ b/db/users/update_password.go
@@ -74,16 +74,16 @@ func UpdateUserPassword(ctx context.Context, app, addonUUID, username string) er
 		if err != nil {
 			return errors.Wrap(ctx, err, "update password of the given database user")
 		}
-	} else {
-		databaseUser, err = c.DatabaseUserResetPassword(ctx, app, addonUUID, username)
-		if err != nil {
-			return errors.Wrap(ctx, err, "reset the password of the given database user")
-		}
 
-		fmt.Printf("User \"%s\" updated with password \"%s\".\n", databaseUser.Name, databaseUser.Password)
+		fmt.Printf("User \"%s\" password updated.\n", databaseUser.Name)
 		return nil
 	}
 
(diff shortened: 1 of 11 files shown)


db/users/update_password.go
```diff
@@ -60,25 +59,31 @@ func UpdateUser(ctx context.Context, app, addonUUID, username string) error {
...
+		userUpdateParam := scalingo.DatabaseUpdateUserParam{
+			DatabaseID:           addonUUID,
+			Password:             password,
+			PasswordConfirmation: confirmedPassword,
+		}
+		databaseUser, err = c.DatabaseUpdateUser(ctx, app, addonUUID, username, userUpdateParam)
+		if err != nil {
+			return errors.Wrap(ctx, err, "update password of the given database user")
+		}
+	} else {
```
EtienneM, line 77: suggestion: I like returning the function as soon as possible. In this case, you could return here and the `else` is useless. There is less indentation which makes the code easier to read IMO
```suggestion
		}
		
		fmt.Printf("User \"%s\" password updated.\n", databaseUser.Name)
		return nil
	}
```

EtienneM (approved):
Just a small suggestion but LGTM :+1: 

Did you ask product about the CLI backward compatibility?

db/users/update_password.go
sc-zenokerr: Thanks. I actually misread the original flow. Your suggestion is absolutely correct. Thanks

sc-zenokerr: > Did you ask product about the CLI backward compatibility?

Yes. My question was answered by benjamin in the comments in the [STORY-642] in notion. 

It should be (remain) possible to set a user-supplied password using the CLI.

EtienneM (approved):

PR #1072, opened 2024-07-08
[STORY-642] Bump version 1.32.1

sc-zenokerr:


sc-zenokerr:
I'm OK with not doing a release, if that is preferred. There is no change to the CLI functionality as the changes are implementation details, (which is why I chose a patch version change, instead of a minor version change). Happy to go with your normal approach.
Please let me know if you think is preferrable
1. minor version change
2. no release

PR #1077, opened 2024-09-16
[STORY-648] feat: raise DB user passwords minimum length to 24

sc-david-voisin:
DB user passwords length must be between 24 to 64.

STORY-648

### DIFF
--- a/db/users/utils_test.go
+++ b/db/users/utils_test.go
@@ -1,0 +1,108 @@
+package users_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/Scalingo/cli/db/users"
+)
+
+func Test_IsPasswordValid(t *testing.T) {
+	testPasswords := map[string]struct {
+		password         string
+		confirmation     string
+		expectedValidity bool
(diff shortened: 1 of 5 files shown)


db/users/utils_test.go
```diff
@@ -0,0 +1,107 @@
+package users_test
```
curzolapierre, line 1: Is there any reason to use a dedicated package for test file?

IMO you can still use `package users` and then not export `IsUsernameValid` function.

curzolapierre (commented):
Just a question/comment otherwise LGTM

db/users/utils_test.go
sc-david-voisin: Good question ;)

Yes, I use a dedicated package for what is called "black-box" testing, that emphasis on user behavior testing. 
More a habit I have from my previous company ^^, and realize that the opposed "white-box" testing suits better for the CLI.

I'll readapt and keep in the same package.

Short discussion black-box vs white-box testing:
https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language/31443271#31443271

PR #1082, opened 2024-10-01
build(asgard): initialize

EtienneM:
I noticed that we didn't initialized this repository with Asgard

=> it does not seem relevant to add a changelog entry for such change


ipfaze (approved):
Nice catch!

PR #1090, opened 2024-12-12
fix(deps): replace deprecated httputil.ClientConn and hijack functionality

### LINKED ISSUE #919: SA1019: httputil.ClientConn has been deprecated since Go 1.0: Use Client or Transport in package net/http instead.
```
apps/run.go:350:12: SA1019: httputil.ClientConn has been deprecated since Go 1.0: Use Client or Transport in package net/http instead. (staticcheck)                                          
        var conn *httputil.ClientConn                                                                                                                                                         
                  ^                                                                            
apps/run.go:356:10: SA1019: httputil.NewClientConn has been deprecated since Go 1.0: Use the Client or Transport in package net/http instead. (staticcheck)                                   
                conn = httputil.NewClientConn(tlsConn, nil)                                                                                                                                   
                       ^                                                                                                                                                                      
apps/run.go:358:10: SA1019: httputil.NewClientConn has been deprecated since Go 1.0: Use the Client or Transport in package net/http instead. (staticcheck)                                   
                conn = httputil.NewClientConn(dial, nil)                                                                                                                                      
```

briceamen:
Fix #919 

Remove usage of deprecated httputil.ClientConn in favor of net/http.Client per golang/go#28030. Since HTTP connection hijacking is no longer supported in modern Go, implemented alternative approach for bidirectional TCP communication.


apps/run.go
```diff
@@ -322,42 +322,43 @@ func (runCtx *runContext) connectToRunServer(ctx context.Context) (*http.Respons
...
-		return nil, nil, errgo.Notef(err, "fail to generate auth")
+		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
-	req.SetBasicAuth("", token)
 
-	url, err := url.Parse(runCtx.attachURL)
+	// Wrap with TLS if using HTTPS
+	if url.Scheme == "https" {
+		tlsConfig := config.TLSConfig.Clone()
+		tlsConfig.ServerName = strings.Split(url.Host, ":")[0]
```
EtienneM, line 339: issue:
```suggestion
		tlsConfig.ServerName = url.Hostname()
```

apps/run.go
```diff
@@ -322,42 +322,43 @@ func (runCtx *runContext) connectToRunServer(ctx context.Context) (*http.Respons
...
 	if err != nil {
 		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
-	token, err := runCtx.scalingoClient.GetAccessToken(ctx)
 
+	// Establish an initial TCP connection
+	conn, err := net.Dial("tcp", url.Host)
 	if err != nil {
-		return nil, nil, errgo.Notef(err, "fail to generate auth")
+		return nil, nil, errgo.Mask(err, errgo.Any)
```
EtienneM, line 333: issue: we are in the process of removing `errgo` from our codebase. When you see a use of `errgo`, please replace it with the use of `github.com/Scalingo/go-utils/errors/v2` (https://www.notion.so/scalingooriginal/Go-Error-Management-1ca42114c730415e95755027935eaccc)
```suggestion
		return nil, nil, errors.Wrap(ctx, err, "establish initial TCP connection")
```

apps/run.go
```diff
@@ -322,42 +322,43 @@ func (runCtx *runContext) connectToRunServer(ctx context.Context) (*http.Respons
...
+	req, err := http.NewRequest("CONNECT", runCtx.attachURL, nil)
 	if err != nil {
 		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
 
-	dial, err := net.Dial("tcp", url.Host)
+	token, err := runCtx.scalingoClient.GetAccessToken(ctx)
 	if err != nil {
-		return nil, nil, errgo.Mask(err, errgo.Any)
+		return nil, nil, errgo.Notef(err, "fail to generate auth")
```
EtienneM, line 350: issue: another thing which changed about our error message is that we no longer want to include words such as "fail to" in the wrapping error message
```suggestion
		return nil, nil, errors.Wrap(ctx, err, "get access token")
```

apps/run.go
```diff
@@ -322,42 +322,43 @@ func (runCtx *runContext) connectToRunServer(ctx context.Context) (*http.Respons
...
-		tlsConn := tls.Client(dial, tlsConfig)
-		conn = httputil.NewClientConn(tlsConn, nil)
-	} else if url.Scheme == "http" {
-		conn = httputil.NewClientConn(dial, nil)
-	} else {
-		return nil, nil, errgo.Newf("invalid scheme format %s", url.Scheme)
+	// Write the HTTP request directly to the connection
+	err = req.Write(conn)
+	if err != nil {
+		return nil, nil, errgo.Mask(err, errgo.Any)
```
EtienneM, line 357: issue: replace use of `errgo`

apps/run.go
```diff
@@ -371,8 +372,11 @@ func (runCtx *runContext) connectToRunServer(ctx context.Context) (*http.Respons
 		return nil, nil, errgo.Mask(err, errgo.Any)
 	}
 
-	connection, _ := conn.Hijack()
-	return res, connection, nil
+	if resp.StatusCode != http.StatusOK {
+		return resp, nil, nil
+	}
```
EtienneM, line 377: suggestion: I think it should return an error in this case, what do you think?

EtienneM (changes requested):
sorry for the delay to review this PR.

praise: that's a long standing issue, thanks a lot for taking it!

question: you mention a breaking change in the PR description but I don't really understand what it this breaking change. How is it breaking the current CLI API?

question: did you tested your change against the staging and / or prod environment?

briceamen:
> question: you mention a breaking change in the PR description but I don't really understand what it this breaking change. How is it breaking the current CLI API?

I mentioned a breaking change because I'm replacing the use of httputil.ClientConn with net/http.Client, and there's no longer any hijacking properly speaking. I define my raw TCP connection early in the function and then add the HTTP protocol (followed by the TLS layer). Since I couldn't perform unit testing due to an issue with importing the RunService mock, I preferred to raise this as a warning.

> question: did you tested your change against the staging and / or prod environment?

I've tested locally (http) and on staging for the TLS layer. I've not tested against production but in theory I could indeed

apps/run.go
briceamen: I wasn't 100% sure about that part, because I've encountered some 202 while testing on staging, and 202 is not 200 (yet).

What could be the best solution ? add a retry and timeout if 2O2 and for everything else that is not 200 or 202 return an error ?

apps/run.go
briceamen: I'll refactor run.go and submit a proper commit for this

briceamen:
Let me know what you think about those changes, I took the opportunity to replace errgo throughout all errors.

EtienneM (approved):

PR #1117, opened 2025-06-20
fix: typo in notifiers example

VincentHardouin:
- [ ] fix typo in notifiers example


EtienneM (approved):
nice catch, thanks a lot :)

PR #1118, opened 2025-06-27
[STORY-1429] chore(deps): Bump go-scalingo to v8.1.0

aurelien-reeves-scalingo:
go-scalingo v8.1.0 add support for limited collaborators

However this PR does not bring support for limited collaborators to the
CLI. It just update Collaborator.Add to add non-limited collaborator in
a retro-compatible way.

Full support for limited collaborator in the CLI is plan for later


go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	dario.cat/mergo v1.0.2 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/ProtonMail/go-crypto v1.3.0 // indirect
+	github.com/Scalingo/go-scalingo/v8 v8.1.0 // indirect
```
EtienneM: issue: this dependency should not be an "indirect" one. Did you execute both `go mod vendor` and `go mod tidy`?

collaborators/add.go
```diff
@@ -7,14 +7,15 @@ import (
 
 	"github.com/Scalingo/cli/config"
 	"github.com/Scalingo/cli/io"
+	"github.com/Scalingo/go-scalingo/v8"
 )
 
-func Add(ctx context.Context, app, email string) error {
+func Add(ctx context.Context, app, email string, isLimited bool) error {
```
EtienneM: suggestion: when calling `collaborators.Add` in `cmd/collaborators.go`, I find it hard to understand what is the last argument. What do you think about using `scalingo.CollaboratorAddParams` here?

```suggestion
func Add(ctx context.Context, app, params scalingo.CollaboratorAddParams) error {
```

go.mod
aurelien-reeves-scalingo: indeed

fixed, thanks

collaborators/add.go
aurelien-reeves-scalingo: Indeed

Done :+1:

PR #1121, opened 2025-07-03
[STORY-2189] Warn user before unscheduling periodic backups

matthieu526-scalingo:
[STORY-2189](https://www.notion.so/scalingooriginal/FEAT-Warn-the-user-when-he-she-disables-an-automated-backup-CLI-225f536d89bf80c6b050c2dd86922a9e)

This story implements a new check to warn a user before unscheduling periodic backups. The user will now be prompted to confirm, or the command will fail.

I found [this](https://github.com/Scalingo/cli/blob/0856226f2ebbb73b49c10eb270f7aa8218b497bb/cmd/log_drains.go#L221) function that is used elsewhere to do the same thing.

Here are things that were done as well:
- Aligning several methods by using a `BoolPtr` function
- Refactoring `askContinue` to be more generic

State of the QA:
- [ ] Done

### DIFF
--- a/cmd/databases.go
+++ b/cmd/databases.go
@@ -115,6 +115,12 @@ var (
 			}
 
 			if disable {
+				continueB := askContinue("This operation will disable the periodic backups of a database. Are you sure?")
+				if !continueB {
+					errorQuit(c.Context, errors.New("unschedule operation was not confirmed"))
+					return nil
+				}
+
 				params.Enabled = utils.BoolPtr(false)
 			}
 			if scheduleAtFlag != "" {
--- a/cmd/databases.go
+++ b/cmd/databases.go
@@ -115,12 +115,10 @@ var (
 			}
 
 			if disable {
-				f := false
-				params.Enabled = &f
+				params.Enabled = utils.BoolPtr(false)
 			}
 			if scheduleAtFlag != "" {
-				t := true
-				params.Enabled = &t
+				params.Enabled = utils.BoolPtr(true)
 				scheduleAt, loc, err := parseScheduleAtFlag(scheduleAtFlag)
 				if err != nil {
 					errorQuit(c.Context, err)
--- a/cmd/alerts.go
+++ b/cmd/alerts.go
@@ -205,9 +205,8 @@ var (
 
 			utils.CheckForConsent(c.Context, currentApp, utils.ConsentTypeContainers)
 
-			disabled := false
 			err := alerts.Update(c.Context, currentApp, c.Args().First(), scalingo.AlertUpdateParams{
-				Disabled: &disabled,
+				Disabled: utils.BoolPtr(false),
 			})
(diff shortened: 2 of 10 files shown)


cmd/databases.go
```diff
@@ -115,12 +115,16 @@ var (
 			}
 
 			if disable {
-				f := false
-				params.Enabled = &f
+				continueB := askContinue("This operation will disable the periodic backups of a database. Are you sure?")
```
leo-scalingo, line 118: Our CLI should be runnable without any interaction to be used in script.

We are using the `DISABLE_INTERACTIVE` environment variable in other places which skips this kind of confirmation, I think it would be good to use it.

```
config/config.go
31:	DisableInteractive   bool   `envconfig:"DISABLE_INTERACTIVE"`
```

(Other pattern is having a --force flag , but it's not a pattern we've been using so far)

matthieu526-scalingo: Is it really the case? From the two places I found as example, that property isn't handled :smile: .
- [Here](https://github.com/Scalingo/cli/blob/e9750c6cdade24d7cf520003a09d17ff43dd91ac/cmd/log_drains.go#L199-199)
- And [here](https://github.com/Scalingo/cli/blob/e9750c6cdade24d7cf520003a09d17ff43dd91ac/apps/destroy.go#L31-L31)
And actually that second example does use the `force` flag [here](https://github.com/Scalingo/cli/blob/e9750c6cdade24d7cf520003a09d17ff43dd91ac/cmd/destroy.go#L17-L17).

So, just to be sure I got this right, I should check `config.C.DisableInteractive` in my code and react accordingly? Meaning blocking this action in a script?

leo-scalingo: :facepalm: 

Indeed and maybe a common `askConfirmation` should be created reading this property from the configuration.

I'm wondering if we should differenciate
* Actions asking for interaction as input params/confirmation  (skippable through env)
* Destructive actions (flag)

matthieu526-scalingo: I don't disagree, and a common `askConfirmation` taking into account the property could be created. But I believe this is slightly out of scope, as here we want to get rid of an NC, and the refactoring would required identifying all places where user input is required.

I also agree with the second point, it could be super interesting to have the combo: force flag + skip interaction through env, handled everywhere the same way.

leo-scalingo: Should we create a ticket dedicated to that and validate this PR then?

matthieu526-scalingo: That sound good to me, at least for once I've done the QA.

leo-scalingo: https://github.com/Scalingo/cli/issues/1122

leo-scalingo (approved):
After our conversation, actually LGTM

sc-david-voisin (approved):

PR #1131, opened 2025-07-29
[STORY-2325] Update applications commands to include projects

matthieu526-scalingo:
This PR updates the CLI to add the project attributes in the applications. It also creates a new filter by `project`.

- [X] QA on staging
- [x] Update to the latest go-scalingo [once merged](https://github.com/Scalingo/go-scalingo/pull/419)


cmd/apps.go
```diff
@@ -1,27 +1,38 @@
...
 	"github.com/Scalingo/cli/detect"
+	"github.com/Scalingo/go-utils/errors/v2"
 )
 
 var (
 	appsCommand = cli.Command{
 		Name:        "apps",
 		Category:    "Global",
 		Description: "List your apps and give some details about them",
+		Flags:       []cli.Flag{&cli.StringFlag{Name: "project", Usage: "Filter apps by project. The filter needs to use the format <ownerUsername>/<projectName>"}},
```
curzolapierre: **Nitpick**: `needs to` is not mandatory
```suggestion
		Flags:       []cli.Flag{&cli.StringFlag{Name: "project", Usage: "Filter apps by project. The filter uses the format <ownerUsername>/<projectName>"}},
```

apps/list.go
```diff
@@ -29,21 +34,27 @@ func List(ctx context.Context) error {
 	}
 
 	t := tablewriter.NewWriter(os.Stdout)
-	t.Header([]string{"Name", "Role", "Status"})
+	t.Header([]string{"Name", "Role", "Status", "Project"})
```
curzolapierre, line 37: **Question**: should we still print an additional column even if there is only one project?
Same question for the case where there is only one project and no project mentioned?

I mean for user PoV that would never use project feature, the `Project` column will always `default`, right?
Is it intended?

curzolapierre (commented):
Some questions, otherwise LGTM

apps/list.go
matthieu526-scalingo: Good question. I implemented what is specified [here](https://www.notion.so/scalingooriginal/IST-Squared-projects-through-the-CLI-terraform-239f536d89bf8092b318efbec6c0c526?source=copy_link#239f536d89bf8092a27dc14c9d427f0e).
But to be honest I think the overhead of parsing the list first to determine the number of columns is maybe a bit overkill? Also, I think a static number of columns is better for a CLI.
I can still ask product.

PS: asked [here](https://scalingo.slack.com/archives/C08DNGCJT3Q/p1753784053148969)

cmd/apps.go
matthieu526-scalingo: Thanks, that's better.

apps/list.go
curzolapierre: > Also, I think a static number of columns is better for a CLI.

Yes I agree, especially because we don't provide a formatted output..

PR #1132, opened 2025-07-31
[STORY-2324] Add all `projects` commands

matthieu526-scalingo:
This PR adds all commands to handle projects.
It is based on top of [this](https://github.com/Scalingo/cli/pull/1131) PR, to avoid some conflicts (dependency to go-scalingo, changelog). Once the base PR is merged, there will just one conflict to use `go-scalingo` in the tagged version.

- [X] QA on staging (which showed several bugs on the API)


cmd/projects.go
```diff
@@ -0,0 +1,141 @@
...
+		}.Render(),
+		Action: func(c *cli.Context) error {
+			projectID := c.Args().First()
+			if projectID == "" {
+				errorQuitWithHelpMessage(errors.New(c.Context, "missing project ID parameter"), c, "projects-update")
+			}
+
+			var newProjectNamePtr *string
+			// Only set a new project name if a new name was provided
+			if newProjectName := c.String("name"); newProjectName != "" {
```
briceamen, line 85: 👀

matthieu526-scalingo: Can you elaborate a bit more :pray: ? :eyes: is not enough to act on :smile: 
Is it because of the short if version?

briceamen: This was a reference to our talk about in line syntax, nothing much to say about it ahah

cmd/autocomplete/projects_list.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/urfave/cli/v2"
+
+	"github.com/Scalingo/cli/config"
+	"github.com/Scalingo/go-utils/errors/v2"
+)
+
+func ProjectsGenericListAutoComplete(c *cli.Context) error {
+	client, err := config.ScalingoClient(c.Context)
+	if err != nil {
+		return errors.Wrap(c.Context, err, "fail to get Scalingo client")
```
briceamen: Nitpick about the fail to

briceamen (approved):
Just a nitpick otherwise LGTM

cmd/autocomplete/projects_list.go
matthieu526-scalingo: Fixed 31b92cdb447f518bc50876225350ed1abd90cef6

PR #1144, opened 2025-09-11
feat(private-networks): list app private network domain names

sc-zenokerr:

Related to [STORY-2435]

- [x] Depends on https://github.com/Scalingo/api/pull/2820
- [ ] Depends on a release of go-scalingo that includes PR https://github.com/Scalingo/go-scalingo/pull/423
- [ ] Update go.mod to the correct release version of go-scalingo

### DIFF
--- a/privatenetworks/list.go
+++ b/privatenetworks/list.go
@@ -58,10 +58,7 @@ func List(ctx context.Context, app string, format string, page uint, perPage uin
 
 		fmt.Printf("Page %d/%d\nTotal number of domains %d\n", domainNames.Meta.CurrentPage, domainNames.Meta.TotalPages, domainNames.Meta.TotalCount)
 	case "json":
-		jsonOutput, err := json.Marshal(domainNames)
-		if err != nil {
-			return errors.Wrapf(ctx, err, "generate output JSON from domain names")
-		}
+		jsonOutput, _ := json.Marshal(domainNames)
 		fmt.Println(string(jsonOutput))
 	}
 


privatenetworks/list.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/Scalingo/cli/config"
+)
+
+const containerTypeWeb = "web"
+
+func List(ctx context.Context, app string, format string, page string, perPage string) error {
+	c, err := config.ScalingoClient(ctx)
+	if err != nil {
+		return errgo.Notef(err, "fail to get Scalingo client")
```
EtienneM: issue: please don't use `errgo`
issue: do not add "fail to" in wrapping error message

privatenetworks/list.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/olekukonko/tablewriter"
+	"gopkg.in/errgo.v1"
+
+	"github.com/Scalingo/cli/config"
+)
+
+const containerTypeWeb = "web"
+
+func List(ctx context.Context, app string, format string, page string, perPage string) error {
+	c, err := config.ScalingoClient(ctx)
```
EtienneM: suggestion: `c` is not a really clear name

```suggestion
  scalingoClient, err := config.ScalingoClient(ctx)
```

cmd/privatenetworks.go
```diff
@@ -0,0 +1,74 @@
...
+				err = privatenetworks.List(ctx, currentApp, formatStr, pageStr, perPageStr)
+			} else {
+				_ = cli.ShowCommandHelp(ctx, c, "private-networks-domain-names")
+			}
+
+			if err != nil {
+				errorQuit(ctx, err)
+			}
+			return nil
+		},
```
EtienneM, line 91: suggestion: I would check the arguments as early as possible and leave the function if needed
```suggestion
		Action: func(ctx context.Context, c *cli.Command) error {
			if c.Args().Len() > 0 {
				err := cli.ShowCommandHelp(ctx, c, "domains")
				if err != nil {
					errorQuit(ctx, err)
				}
				return nil
			}

			pageStr := c.String("page")
			perPageStr := c.String("per-page")
			formatStr := c.String("format")
			ctx, _ = logger.WithFieldsToCtx(ctx, logrus.Fields{
				"page":     pageStr,
				"per_page": perPageStr,
				"format":   formatStr,
			})

			currentApp := detect.CurrentApp(c)
			err := privatenetworks.List(ctx, currentApp, formatStr, pageStr, perPageStr)
			if err != nil {
				errorQuit(ctx, err)
			}
			return nil
		},
```

privatenetworks/list.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	if len(domainNames) == 0 {
+		fmt.Println("No private network domain found")
+		return nil
+	}
+
+	switch format {
+	case "table":
+		sort.Slice(domainNames, func(i, j int) bool {
```
EtienneM: question: could you add a comment to explain how you sort?

suggestion: I would use a named function rather than an anonymous one to shorten the `List` function

privatenetworks/list.go
```diff
@@ -0,0 +1,94 @@
...
+
+		for _, domain := range domainNames {
+			containerType := containerTypeWeb
+			parts := strings.Split(string(domain), ".")
+			slices.Reverse(parts)
+			if len(parts) == 5 {
+				containerType = containerTypeWeb
+			} else if len(parts) > 5 {
+				containerType = parts[5]
+			}
```
EtienneM, line 47: issue: can you add a comment to explain this? That's hard to understand as is

privatenetworks/list.go
sc-zenokerr: Done in https://github.com/Scalingo/cli/pull/1144/commits/2f5a05e0c8bb958265c41b06eafca65f437e38b8

privatenetworks/list.go
sc-zenokerr: Sorting is removed in https://github.com/Scalingo/cli/pull/1144/commits/2f5a05e0c8bb958265c41b06eafca65f437e38b8

cmd/privatenetworks.go
sc-zenokerr: Done in https://github.com/Scalingo/cli/pull/1144/commits/2f5a05e0c8bb958265c41b06eafca65f437e38b8

privatenetworks/list.go
sc-zenokerr: Sorting is removed. The list is returned from the api already sorted.

privatenetworks/list.go
sc-zenokerr: Done in https://github.com/Scalingo/cli/pull/1144/commits/589e2662dfa40146ae9ec724000a8e5324b654a3

privatenetworks/list.go
sc-zenokerr: Done in https://github.com/Scalingo/cli/pull/1144/commits/ca5640642f46cd9bff64fcd2a9cac508e502bce0

cmd/privatenetworks.go
```diff
@@ -0,0 +1,92 @@
...
+			&cli.StringFlag{
+				Name:  "page",
+				Value: "1",
+				Usage: "[page]",
+			},
+			&cli.StringFlag{
+				Name:  "per-page",
+				Value: "20",
+				Usage: "[per-page]",
+			},
```
EtienneM, line 41: suggestion: I can see that we start having lots of paginated commands. Shouldn't we refactor a bit so that we declare the pagination flag in a single place? That would prevent having a pagination with different defaults depending on the command

cmd/privatenetworks.go
EtienneM: The linked commit is not the content of my suggestion. My suggestion was more about `c.Args().Len()` check.

cmd/privatenetworks.go
sc-zenokerr: That's a good idea. I think it would warrant it's own PR. I've created an issue for it. https://github.com/Scalingo/cli/issues/1147

PR #1145, opened 2025-09-17
feat: add database flags on current commands

curzolapierre:
- [x] update go-scalingo dependency once https://github.com/Scalingo/go-scalingo/pull/425 has been merge

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -54,10 +54,8 @@ func addonUUIDFromFlags(ctx context.Context, c *cli.Command, app string, exitIfM
 		addonName = os.Getenv("SCALINGO_ADDON")
 	}
 
-	if addonName == "" {
-		if os.Getenv("SCALINGO_PREVIEW_FEATURES") == "true" {
-			_, addonName = detect.GetCurrentResourceAndDatabase(ctx, c)
-		}
+	if addonName == "" && os.Getenv("SCALINGO_PREVIEW_FEATURES") == "true" {
+		_, addonName = detect.GetCurrentResourceAndDatabase(ctx, c)
 	}
 
 	if addonName == "" && len(exitIfMissing) > 0 && exitIfMissing[0] {
--- a/detect/app.go
+++ b/detect/app.go
@@ -2,7 +2,6 @@ package detect
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"strings"
 
@@ -11,6 +10,7 @@ import (
 	"gopkg.in/errgo.v1"
 
 	"github.com/Scalingo/cli/config"
+	"github.com/Scalingo/cli/io"
 	"github.com/Scalingo/cli/utils"
 	"github.com/Scalingo/go-scalingo/v8/debug"
 	"github.com/Scalingo/go-utils/errors/v2"
@@ -35,7 +35,7 @@ func GetCurrentResourceAndDatabase(ctx context.Context, c *cli.Command) (string,
 	currentApp := extractAppName(c)
 	currentDatabase, databaseUUID, err := currentDatabaseNameAndUUID(ctx, c)
 	if err != nil && !errors.Is(err, errDatabaseNotFound) {
-		fmt.Println(err)
+		io.Error(err)
 		os.Exit(1)
 	}
 
@@ -49,12 +49,12 @@ func GetCurrentResourceAndDatabase(ctx context.Context, c *cli.Command) (string,
 	}
 
 	if appFlagSet && currentDatabase != "" && databaseUUID != "" {
-		fmt.Println("You can't use --app and --database flags together.")
+		io.Error("You can't use --app and --database flags together.")
 		os.Exit(1)
 	}
 
 	if currentApp == "" && currentDatabase == "" {
-		fmt.Println("No application or database found. Please use --app or --database flag.")
+		io.Error("No application or database found. Please use --app or --database flag.")
 		os.Exit(1)
 	}
 
@@ -78,7 +78,7 @@ func CurrentApp(c *cli.Command) string {
 	appName := extractAppName(c)
 
 	if appName == "" {
-		fmt.Println("Unable to find the application name, please use --app flag.")
+		io.Error("Unable to find the application name, please use --app flag.")
 		os.Exit(1)
 	}
 	debug.Println("[detect] App name is", appName)
@@ -110,19 +110,19 @@ func currentDatabaseNameAndUUID(ctx context.Context, c *cli.Command) (string, st
 	}
 
 	if len(addons) == 0 {
-		fmt.Println("No database found with the given name.")
+		io.Warning("No database found with the given name.")
 		return "", "", errDatabaseNotFound
 	}
 
 	if len(addons) > 1 {
-		fmt.Println("Multiple databases found with the given name, it may be an application. Please use the database name instead.")
+		io.Error("Multiple databases found with the given name, it may be an application. Please use the database name instead.")
 		return "", "", errDatabaseNotFound
 	}
 
 	databaseUUID := addons[0].ID
 
 	if databaseUUID == "" {
-		fmt.Println("Unable to find the database ID with the given name.")
+		io.Error("Unable to find the database ID with the given name.")
 		return "", "", errDatabaseNotFound
 	}
 


cmd/flags.go
```diff
@@ -25,6 +25,17 @@ var (
...
+	if os.Getenv("SCALINGO_PREVIEW_FEATURES") == "true" {
+		return &cli.StringFlag{
+			Name:  "database",
+			Value: "<database_name>",
+			Usage: "ID of the current database",
+		}
+	}
+	return &cli.StringFlag{}
+}
+
```
sc-david-voisin, line 39: why embed `cli.StringFlag` creation in a function ?

cmd/log_drains.go
```diff
@@ -170,40 +184,51 @@ Warning: At the moment, only databases addons are able to forward logs to a drai
...
 			}
 			drain := c.Args().First()
 
-			addonID := addonUUIDFromFlags(ctx, c, currentApp)
+			var addonID string
+			if currentDatabase != "" {
+				addonID = currentDatabase
+			} else {
+				addonID = addonUUIDFromFlags(ctx, c, currentResource)
+			}
```
sc-david-voisin, line 202: What about have `addonUUIDFromFlags` itself returns database ID if present ?

cmd/log_drains.go
```diff
@@ -106,11 +114,16 @@ Warning: At the moment, only databases addons are able to forward logs to a drai
 		}.Render(),
 
 		Action: func(ctx context.Context, c *cli.Command) error {
-			currentApp := detect.CurrentApp(c)
+			currentResource, currentDatabase := detect.GetCurrentResourceAndDatabase(ctx, c)
 
-			addonID := addonUUIDFromFlags(ctx, c, currentApp)
+			utils.CheckForConsent(ctx, currentResource)
```
sc-david-voisin, line 121: As I understand, currentResource can be either an app or a database. But can database always substitute to the app ?
What happen in this case with the CheckConsent ?

cmd/flags.go
```diff
@@ -25,6 +25,17 @@ var (
 	}
 )
 
+func databaseFlag() *cli.StringFlag {
+	if os.Getenv("SCALINGO_PREVIEW_FEATURES") == "true" {
```
matthieu526-scalingo, line 30: question: should the flag be generic to previews, or specific to the DB preview?

cmd/flags.go
```diff
@@ -25,6 +25,17 @@ var (
...
+		return &cli.StringFlag{
+			Name:  "database",
+			Value: "<database_name>",
+			Usage: "ID of the current database",
+		}
+	}
+	return &cli.StringFlag{}
```
matthieu526-scalingo, line 37: question: should it return empty flag or `nil` as this is a pointer?

cmd/log_drains.go
```diff
@@ -38,17 +39,23 @@ Use the parameter "--with-addons" to list log drains of all addons connected to
...
-			utils.CheckForConsent(ctx, currentApp)
+			utils.CheckForConsent(ctx, currentResource)
 
-			addonID := addonUUIDFromFlags(ctx, c, currentApp)
+			var addonID string
+			if currentDatabase != "" {
+				addonID = currentDatabase
+			} else {
+				addonID = addonUUIDFromFlags(ctx, c, currentResource)
+			}
```
matthieu526-scalingo: It could be simplified.

```suggestion
			addonID = currentDatabase
			if currentDatabase == "" {
				addonID = addonUUIDFromFlags(ctx, c, currentResource)
			}
```

logdrains/list.go
```diff
@@ -23,7 +23,7 @@ type ListAddonOpts struct {
 	AddonID    string
 }
 
-func List(ctx context.Context, app string, opts ListAddonOpts) error {
+func List(ctx context.Context, resource string, opts ListAddonOpts) error {
```
matthieu526-scalingo:
```suggestion
func List(ctx context.Context, resourceID string, opts ListAddonOpts) error {
```

cmd/flags.go
curzolapierre: Because it's needed by the urfave/cli in order to parse flags.
If you meant to replace this by `nil` it'd panic instead.

cmd/log_drains.go
curzolapierre: I was waiting for your change on go-scalingo :grin: 
cf: a90aaf68

cmd/log_drains.go
curzolapierre: Yes I'm wondering too.
It'd bring some magic to avoid to change command's code but may hide some logic.
WDYT?

config/config.go
```diff
@@ -219,6 +219,23 @@ func ScalingoClient(ctx context.Context) (*scalingo.Client, error) {
 	return scalingo.New(ctx, config)
 }
 
+func ScalingoDatabaseClient(ctx context.Context) (*scalingo.PreviewClient, error) {
+	authenticator := &CliAuthenticator{}
+	_, token, err := authenticator.LoadAuth(ctx)
+	if err != nil {
+		return nil, errgo.Notef(err, "fail to load credentials")
```
matthieu526-scalingo: Can you please update to the "latest" error management?

cmd/log_drains.go
```diff
@@ -38,17 +39,23 @@ Use the parameter "--with-addons" to list log drains of all addons connected to
...
-			currentApp := detect.CurrentApp(c)
+			currentResource, currentDatabase := detect.GetCurrentResourceAndDatabase(ctx, c)
+
 			if c.Args().Len() != 0 {
 				_ = cli.ShowCommandHelp(ctx, c, "log-drains")
 				return nil
 			}
 
-			utils.CheckForConsent(ctx, currentApp)
+			utils.CheckForConsent(ctx, currentResource)
```
matthieu526-scalingo: I wonder if the signature of `CheckForConsent` should be updated to already tell if we target a DB Or an app?

utils/consent.go
```diff
@@ -69,6 +69,26 @@ func CheckForConsent(ctx context.Context, appName string, consentTypes ...Consen
 		}
 	}
 
+	dbClient, err := config.ScalingoDatabaseClient(ctx)
```
matthieu526-scalingo, line 73: Will this work? Especially after the apps, if you don't have the consent, you'll fallback to DBs. Is this what we want?

logdrains/add.go
```diff
@@ -16,12 +16,22 @@ type AddDrainOpts struct {
...
+func Add(ctx context.Context, app, database string, opts AddDrainOpts) error {
 	c, err := config.ScalingoClient(ctx)
 	if err != nil {
 		return errgo.Notef(err, "fail to get Scalingo client to add a log drain")
 	}
 
+	if database != "" {
+		d, err := c.LogDrainAddonAdd(ctx, app, database, opts.Params)
+		if err != nil {
+			io.Status("fail to add drain to", "'"+app+"' database addon:\n\t", err)
```
matthieu526-scalingo, line 28: suggestion; this could (of should?) be a warning/

detect/app.go
```diff
@@ -1,38 +1,78 @@
...
+//
+// GetCurrentResourceAndDatabase returns the current resource (app or database)
+// and the current database UUID if any.
+// It exits CLI in case of error.
+func GetCurrentResourceAndDatabase(ctx context.Context, c *cli.Command) (string, string) {
+	if os.Getenv("SCALINGO_PREVIEW_FEATURES") != "true" {
+		return CurrentApp(c), ""
+	}
+
+	currentApp := extractAppNameFromCommandLineOrEnvOrGit(c)
```
matthieu526-scalingo: No need to complexify with "or or or" IMO :slightly_smiling_face: 
```suggestion
	currentApp := extractAppName(c)
```

detect/app.go
```diff
@@ -42,6 +82,46 @@ func CurrentApp(c *cli.Command) string {
...
+
+	if len(addons) == 0 {
+		fmt.Println("Unable to find the database with the given name.")
+		return "", "", errDatabaseNotFound
+	}
+
+	databaseUUID := addons[0].ID
+
+	if databaseUUID == "" {
+		fmt.Println("Unable to find the database with the given name.")
```
matthieu526-scalingo: suggestion: I would change the error message so the 2 are different. It will help debug properly.

cmd/flags.go
matthieu526-scalingo: Sorry I though this was answered in another comment.

cmd/log_drains.go
sc-david-voisin: First, both solutions suit me.
Then, is that such a problem to have implicit logic in addonUUIDFromFlags ? The function is quickly read. Eventually, a comment may be added to make it more explicit.

cmd/flags.go
curzolapierre: We don't have other kind of preview features that's why I stuck with what had been proposed in the story.

cmd/flags.go
curzolapierre: Yes it was answered here: https://github.com/Scalingo/cli/pull/1145#discussion_r2357768454

logdrains/add.go
curzolapierre: Some addons don't support log drains so I think an simple `Info` is enough.

logdrains/list.go
curzolapierre: It'd be more `resourceName` than `resourceID`

cmd/log_drains.go
curzolapierre: Probably yes, handled [here](https://github.com/Scalingo/cli/compare/a90aaf68dc87551cca294ee044ea69fcbcc91d1b..40937ef9d0867f1649ccdf8b90689c134e9e857c#diff-667e40be90d47d4e0e34cdf2382cc62bdc39b2e3620b118412d4938acc97e43cR22-R94)

utils/consent.go
curzolapierre: Here I only fallback for DBs to check if the admin is a collaborator of the given database. So yes it is what we want IMO.

But something was missing: `consentTypes`. Because in case of database we have to look only for DB consent.

So again same question as before: should the helper handle the case by itself or it'd be the reponsability of the caller?
For example [here](https://github.com/Scalingo/cli/compare/a90aaf68dc87551cca294ee044ea69fcbcc91d1b..40937ef9d0867f1649ccdf8b90689c134e9e857c#diff-b5e15574910225e625c844a3b06124ee00e735d9c5f00a47684e94751b8e8284R49-R55)

detect/app.go
curzolapierre: Done [here](https://github.com/Scalingo/cli/compare/a90aaf68dc87551cca294ee044ea69fcbcc91d1b..40937ef9d0867f1649ccdf8b90689c134e9e857c#diff-42d00f5c22847004fbd52671050c6c5b843b515fae446aa59f2026e67a157d2cR116-R120)

config/config.go
curzolapierre: Done [here](https://github.com/Scalingo/cli/compare/a90aaf68dc87551cca294ee044ea69fcbcc91d1b..40937ef9d0867f1649ccdf8b90689c134e9e857c#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R227-R235)

detect/app.go
curzolapierre: Done [here](https://github.com/Scalingo/cli/compare/a90aaf68dc87551cca294ee044ea69fcbcc91d1b..40937ef9d0867f1649ccdf8b90689c134e9e857c#diff-42d00f5c22847004fbd52671050c6c5b843b515fae446aa59f2026e67a157d2cR31-R74)

cmd/collaborators.go
```diff
@@ -112,21 +112,22 @@ var (
 		},
 	}
 
-	CollaboratorsUpdateCommand = cli.Command{
+	collaboratorsUpdateCommand = cli.Command{
```
curzolapierre, line 115: Limited collaborators are not allowed on database:

We can rely on the API error message for that, for example:
```sh
scalingo --database my-db collaborators-update --limited user@example.com
	!     400 Bad Request → cannot change collaborator role for a dedicated database app
```

matthieu526-scalingo:
@curzolapierre just to let you know [this](https://github.com/Scalingo/cli/pull/1148) was merged since and it could impact you.

cmd/log_drains.go
curzolapierre: Ok I gave it a try, here is the result: ba412cf35349516fd7ddbae34554198e83efeb96

sc-david-voisin: It is not bad, reading is improved.

config/config.go
```diff
@@ -219,6 +220,23 @@ func ScalingoClient(ctx context.Context) (*scalingo.Client, error) {
 	return scalingo.New(ctx, config)
 }
 
+func ScalingoDatabaseClient(ctx context.Context) (*scalingo.PreviewClient, error) {
```
sc-david-voisin, line 223: Back to our old discussion ^^, I find this client usage masks the "client.Preview()" call introduced in `go-scalingo`, and I prefer to explicitly use Preview at each call to remember it.

cmd/collaborators.go
```diff
@@ -13,14 +13,14 @@ import (
 )
 
 var (
-	CollaboratorsListCommand = cli.Command{
+	collaboratorsListCommand = cli.Command{
 		Name:        "collaborators",
 		Category:    "Collaborators",
 		Usage:       "List the collaborators of an application",
-		Flags:       []cli.Flag{&appFlag},
+		Flags:       []cli.Flag{&appFlag, databaseFlag()},
```
matthieu526-scalingo, line 20: At first it was a bit confusing to see a variable and a function, but I understand why.

config/config.go
curzolapierre: Actually, I don't think we need to know that this is a preview client here in the CLI.

In addition, it matches `ScalingoClient` helper, so in code we'd see something like:
```
dbClient, err := config.ScalingoDatabaseClient(ctx)
```
it's cristal clear what the purpose of the client. We don't need to know that's a "preview" client for a CLI point of view.

sc-david-voisin: I found interesting to keep the "preview" visible, as we implemented the method Preview on purpose I believe.

But it is not a strong point, so go for `ScalingoDatabaseClient`.

cmd/flags.go
```diff
@@ -40,6 +54,12 @@ func addonUUIDFromFlags(ctx context.Context, c *cli.Command, app string, exitIfM
 		addonName = os.Getenv("SCALINGO_ADDON")
 	}
 
+	if addonName == "" {
+		if os.Getenv("SCALINGO_PREVIEW_FEATURES") == "true" {
+			_, addonName = detect.GetCurrentResourceAndDatabase(ctx, c)
+		}
+	}
+
```
matthieu526-scalingo:
```suggestion
	if addonName == "" && os.Getenv("SCALINGO_PREVIEW_FEATURES") == "true" {
		_, addonName = detect.GetCurrentResourceAndDatabase(ctx, c)
	}

```

cmd/run.go
```diff
@@ -79,7 +80,18 @@ var (
...
-			currentApp := detect.CurrentApp(c)
+			currentResource := detect.GetCurrentResource(ctx, c)
+			isDB, err := utils.IsResourceDatabase(ctx, currentResource)
+			if err != nil && !errors.Is(err, utils.ErrResourceNotFound) {
+				errorQuit(ctx, err)
+			}
+			if isDB {
+				io.Error("It is currently impossible to use `run` on a database.")
+				return nil
+			}
```
matthieu526-scalingo, line 91: suggestion: this is done so many time, that it could be an additional parameter sent to `GetCurrentResource` for example `checkIsDB bool` and could return another bool as well.

detect/app.go
```diff
@@ -1,38 +1,82 @@
...
-			appName = os.Getenv("SCALINGO_APP")
-		} else if dir, ok := utils.DetectGit(); ok {
-			appName, err = GetAppNameFromGitRemote(dir, RemoteNameFromFlags(c))
-			if err != nil {
-				debug.Println(err)
-			}
+	currentApp := extractAppName(c)
+	currentDatabase, databaseUUID, err := currentDatabaseNameAndUUID(ctx, c)
+	if err != nil && !errors.Is(err, errDatabaseNotFound) {
+		fmt.Println(err)
```
matthieu526-scalingo: Should we use the fmt package here, when  we have the io one?

curzolapierre: Indeed, I've forgot about it

cmd/run.go
curzolapierre: I wonder about this. I don't find it really a KISS solution.
Either I don't see how it'd help readability because we'll still need to check if it's a DB or not or I'd prefer to keep these repetitions

detect/app.go
```diff
@@ -1,38 +1,83 @@
...
+}
+
+// CurrentApp returns the app name if it has been found in one of the following:
+// app flag, environment variable "SCALINGO_APP", current Git remote.
+// It returns an empty string if not found.
+func CurrentApp(c *cli.Command) string {
+	appName := extractAppName(c)
+
 	if appName == "" {
 		fmt.Println("Unable to find the application name, please use --app flag.")
```
matthieu526-scalingo: nitpick Leftover fmt.println

matthieu526-scalingo (approved):

PR #1146, opened 2025-09-19
[STORY-2512] List, create, delete databases next generation

sc-david-voisin:
Unleash the Database Next Generation power in your favorite CLI !

### DIFF
--- a/dbng/list.go
+++ b/dbng/list.go
@@ -23,7 +23,7 @@ func List(ctx context.Context) error {
 		return errors.Wrap(ctx, err, "list databases")
 	}
 
-	io.Warning("This command only displays databases next generation where you are the owner")
+	io.Warning("This command only displays databases next generation you own.")
 
 	t := tablewriter.NewWriter(os.Stdout)
 	t.Header([]string{"ID", "Name", "Type", "Plan", "Role", "Status", "Project"})


dbng/list.go
```diff
@@ -0,0 +1,47 @@
...
+func List(ctx context.Context) error {
+	c, err := config.ScalingoClient(ctx)
+	if err != nil {
+		return errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	databases, err := c.Preview().DatabasesList(ctx)
+	if err != nil {
+		return errors.Wrap(ctx, err, "list databases")
+	}
```
curzolapierre, line 24: What about using same kind of helper as `config.ScalingoClient` for database client?
```suggestion
	dbClient, err := config.ScalingoDatabaseClient(ctx)
	if err != nil {
		return errors.Wrap(ctx, err, "get database Scalingo client")
	}
```

sc-david-voisin: I am not convinced by this abstraction that hides we are using a `Preview()`  feature especially introduced in` go-scalingo`.

Your change proposal is wrong I think, I understand it would rather change these two lines:
l16: 	`c, err := config.ScalingoDatabaseClient(ctx)`
l21:          `databases, err := c.DatabasesList(ctx)`

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## To Be Released
 
+* feat(projects): create all commands to handle next generation `databases`
```
matthieu526-scalingo: suggestion: I think the changelog should include the same kind of message as [here](https://github.com/Scalingo/cli/pull/1146/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR9).

addons/provision.go
```diff
@@ -26,7 +25,7 @@ func Provision(ctx context.Context, app, addon, plan string) error {
 		return errgo.Notef(err, "fail to get Scalingo client")
 	}
 
-	planID, err := checkPlanExist(ctx, c, addon, plan)
+	planID, err := utils.CheckPlanExist(ctx, c, addon, plan)
```
matthieu526-scalingo: remark(non-blocking): I don't think this is a great method name. I would expect a `bool` to be returned here, no the plan ID. So it sounds like a `GetPlan` or equivalent.

utils/roles.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/Scalingo/go-scalingo/v8"
+)
+
+const (
+	roleOwner        = "owner"
+	roleCollaborator = "collaborator"
+)
+
+type Role string
```
matthieu526-scalingo, line 12: If a new type is created, shouldn't the two roles above use it?

matthieu526-scalingo: (also I am not sure it brings a lot of value here, as it needs to be casted back to a string when used)

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+		Category:  "Databases",
+		Usage:     "View database next generation",
+		ArgsUsage: "database-id",
+		Description: CommandDescription{
+			Description: "View database next generation detailed informations",
+			Examples:    []string{"scalingo database-info database_id"},
+			SeeAlso:     []string{"databases", "database-create", "database-destroy"},
+		}.Render(),
+		Action: func(ctx context.Context, c *cli.Command) error {
+			currentApp := detect.CurrentApp(c)
```
matthieu526-scalingo, line 48: question: this will conflict with what Pierre is doing, right?

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+
+			err := dbng.Show(ctx, c.Args().First())
+			if err != nil {
+				errorQuit(ctx, err)
+			}
+
+			return nil
+		},
+		ShellComplete: func(_ context.Context, c *cli.Command) {
+			_ = autocomplete.CmdFlagsAutoComplete(c, "database-info")
```
matthieu526-scalingo, line 64: question: would it make sense creating a custom autocomplete? I did one you could see on `projects` (this could be used for show and destroy)

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+			Examples: []string{
+				"scalingo database-create --type postgresql --plan postgresql-dbng-starter-2048 my_super_database",
+				"scalingo database-create --type postgresql --plan postgresql-dbng-starter-2048 --wait my_super_database",
+			},
+			SeeAlso: []string{"databases", "database-info", "database-destroy"},
+		}.Render(),
+		Action: func(ctx context.Context, c *cli.Command) error {
+			currentApp := detect.CurrentApp(c)
+			if c.Args().Len() != 1 {
+				return cli.ShowCommandHelp(ctx, c, "database-create")
```
matthieu526-scalingo, line 92: suggestion: could it make sense to handle such validation as it was done [here](https://github.com/Scalingo/cli/blob/49cc580c939a03d5f47b132a56cefe6a65899ced/cmd/apps.go#L76-L76)?

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/Scalingo/cli/cmd/autocomplete"
+	"github.com/Scalingo/cli/dbng"
+	"github.com/Scalingo/cli/detect"
+	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v8"
+)
+
+var (
+	databasesList = cli.Command{
```
matthieu526-scalingo: nitpick: I think all commands are suffixed `command` no?
```suggestion
	databasesListCommand = cli.Command{
```

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+			}
+
+			return nil
+		},
+		ShellComplete: func(_ context.Context, c *cli.Command) {
+			_ = autocomplete.CmdFlagsAutoComplete(c, "database-info")
+		},
+	}
+
+	databaseAdd = cli.Command{
```
matthieu526-scalingo: nitpick: command is `database-create` but variable is `databaseAdd` :slightly_smiling_face:

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+			},
+			SeeAlso: []string{"databases", "database-info", "database-destroy"},
+		}.Render(),
+		Action: func(ctx context.Context, c *cli.Command) error {
+			currentApp := detect.CurrentApp(c)
+			if c.Args().Len() != 1 {
+				return cli.ShowCommandHelp(ctx, c, "database-create")
+			}
+
+			utils.CheckForConsent(ctx, currentApp)
```
matthieu526-scalingo, line 95: question: do you need to check the consent here? I dont think this is relevant as their is no app here.

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+
+			params := scalingo.DatabaseCreateParams{
+				AddonProviderID: c.String("type"),
+				PlanID:          c.String("plan"),
+				ProjectID:       c.String("project"),
+				Name:            c.Args().First(),
+			}
+			err := dbng.Add(ctx, params, c.Bool("wait"))
```
matthieu526-scalingo, line 103: question: why is `wait` passed outside of `DatabaseCreateParams`?

cmd/databases_ng.go
```diff
@@ -0,0 +1,141 @@
...
+			}
+
+			return nil
+		},
+		ShellComplete: func(_ context.Context, c *cli.Command) {
+			_ = autocomplete.CmdFlagsAutoComplete(c, "database-create")
+		},
+	}
+
+	databaseDelete = cli.Command{
```
matthieu526-scalingo: nitpick: command is `database-destroy` and variable `databaseDelete`

dbng/add.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/Scalingo/cli/config"
+	"github.com/Scalingo/cli/io"
+	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v8"
+	"github.com/Scalingo/go-utils/errors/v2"
+)
+
+const dbngTypePostgresql = "dedicated-pg"
+
+var aliases = map[string]string{
```
matthieu526-scalingo, line 18: question: Do we really want to do this? Meaning allowing a ng DB to be created with the `psql` name? (for example)

cmd/databases_ng.go
matthieu526-scalingo: PS: okay now I understand why, then it makes sense.

dbng/add.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return errors.Wrap(ctx, err, "add database")
+	}
+	io.Statusf("Your %s database %s ('%s') is being provisioned…\n\n",
+		db.Addon.AddonProvider.Name,
+		db.App.ID,
+		db.App.Name,
+	)
+
+	if wait {
```
matthieu526-scalingo, line 56: question: Is this really a behaviour we want to keep? Will people really do that for 15-20 minutes (or more)?

dbng/add.go
```diff
@@ -0,0 +1,89 @@
...
+		spinner := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
+		spinner.Suffix = " Waiting for the database to be ready\n"
+		spinner.Start()
+		defer spinner.Stop()
+
+		err = waitForRunningDatabase(ctx, c, db.App.ID)
```
matthieu526-scalingo, line 62: question: isn't the DB ID returned at the root object? It would be better not to use the `app` whenever possible.

dbng/add.go
```diff
@@ -0,0 +1,89 @@
...
+// waitForRunningDatabase is a blocking function waiting for the database to be running.
+func waitForRunningDatabase(ctx context.Context, client *scalingo.Client, appID string) error {
+	for range time.Tick(10 * time.Second) {
+		db, err := client.Preview().DatabaseShow(ctx, appID)
+		if err != nil {
+			if errors.Is(err, scalingo.ErrDatabaseNotFound) {
+				continue
+			}
+			return err
+		}
```
matthieu526-scalingo: nitpick: i would do an `if else if`.
```suggestion
	if errors.Is(err, scalingo.ErrDatabaseNotFound) {
		continue
	} else if err != nil {
			return err
	}
```

dbng/add.go
```diff
@@ -0,0 +1,89 @@
...
+		if db.Database.Status == scalingo.DatabaseStatusRunning && db.App.Status == scalingo.AppStatusRunning {
```
matthieu526-scalingo, line 83: question: are you sure the "invisible" application status will be `running` as well?

dbng/add.go
```diff
@@ -0,0 +1,89 @@
...
+		spinner.Stop()
+		io.Status("Your database is up and running.")
+	}
+
+	return nil
+}
+
+// waitForRunningDatabase is a blocking function waiting for the database to be running.
+func waitForRunningDatabase(ctx context.Context, client *scalingo.Client, appID string) error {
+	for range time.Tick(10 * time.Second) {
```
matthieu526-scalingo, line 76: question: IF we keep this, don't we want to timeout?

dbng/delete.go
```diff
@@ -0,0 +1,47 @@
...
+	if validationID != db.App.ID && validationID != db.App.Name {
+		return errors.Newf(ctx, "'%s' is not the ID or the name of the database, aborting…\n", validationID)
+	}
+
+	err = c.Preview().DatabaseDestroy(ctx, appID)
+	if err != nil {
+		return errors.Wrap(ctx, err, "delete database")
+	}
+
+	io.Status("The database is being deleted.")
```
matthieu526-scalingo, line 44: question: do we want to display this? Or something like `deletion order is being processed` (as we know this is not instantaneous).

dbng/list.go
```diff
@@ -0,0 +1,55 @@
...
+
+	io.Warning("This command only displays databases next generation where you are the owner")
```
matthieu526-scalingo: question: does this make sense?

dbng/list.go
```diff
@@ -0,0 +1,55 @@
...
+
+	t := tablewriter.NewWriter(os.Stdout)
+	t.Header([]string{"ID", "Name", "Type", "Plan", "Role", "Status", "Project"})
+
+	currentUser, err := config.C.CurrentUser(ctx)
```
matthieu526-scalingo, line 31: question: is this required, if we only display the "owned" databases?

dbng/list.go
```diff
@@ -0,0 +1,55 @@
...
+			db.Addon.AddonProvider.Name,
+			db.Addon.Plan.Name,
+			string(role),
+			string(db.Addon.Status),
+			db.App.Project.Name,
+		})
+	}
+	_ = t.Render()
+
+	io.Info("Looking for database addons attached to your applications? Use the 'addons' command")
```
matthieu526-scalingo, line 51: praise: loving this!

dbng/show.go
```diff
@@ -0,0 +1,53 @@
...
+		{"Type", db.Addon.AddonProvider.Name},
+		{"Plan", db.Addon.Plan.Name},
+		{"Status", string(db.Database.Status)},
+		{"Version", db.Database.ReadableVersion},
+		{"Force TLS  ", forceSSL},
+		{"Internet Accessibility", internetAccess},
+		{"Maintenance window", utils.FormatMaintenanceWindowWithTimezone(db.Database.MaintenanceWindow, time.Local)},
+	}
+	t := tablewriter.NewWriter(os.Stdout)
+	_ = t.Bulk(data)
```
matthieu526-scalingo, line 49: question: What is this used for? I just noticed it seems different from the `list` command.

matthieu526-scalingo (commented):
Sorry for the number of comments :sweat_smile:. There is nothing major, but  a bunch of questions and small suggestions.

addons/provision.go
sc-david-voisin: I picked an existing one, I'll rename.

cmd/databases_ng.go
sc-david-voisin: Yes, a conflict is expected with @curzolapierre PR here.
-> `CurrentApp` aims to be replaced by `GetCurrentResource`

cmd/databases_ng.go
sc-david-voisin: I know! I wanted to keep coherence with the name file add.go that exist for other CRUD entities, but the overall project lacks coherence in fine.
No strong opinion, I can do, in such case I rename the files as well and I have at least the same case with `dabataseShow` for database-info.

cmd/databases_ng.go
sc-david-voisin: I have a look.

cmd/databases_ng.go
sc-david-voisin: To me, they are very close, the one for projectID excludes straight away `''` argument.

cmd/databases_ng.go
sc-david-voisin: I've put consent to all DBNG features, but don't know what we want to protect or not.
For the app, a DBNG is at the same time a database and an app, and I believe it is worth for its content.
Eventaully, I am interested by @curzolapierre opinion on this point.

dbng/add.go
sc-david-voisin: I tried to align with what we already have for the addons:
https://github.com/Scalingo/cli/blob/master/utils/addon_uuid.go#L23

dbng/add.go
sc-david-voisin: I agree 15-20 min is long, but this is an optional behavior, asked in the current specs:
https://www.notion.so/scalingooriginal/FEAT-CLI-List-and-create-and-update-database-261f536d89bf80a3a83ee588c1f94d89?source=copy_link#264f536d89bf80ba9278c79da92ec7ce

Looking in the CLI there are other long operations with similar waiting features, such as backup create.

dbng/add.go
sc-david-voisin: Actually the App ID **is** the Database ID.

See comments in this specs block:
https://www.notion.so/scalingooriginal/FEAT-CLI-List-and-create-and-update-database-261f536d89bf80a3a83ee588c1f94d89?source=copy_link#261f536d89bf81058886de1501948745

dbng/add.go
sc-david-voisin: That would be timeout around 1h, is it worth ?

dbng/add.go
sc-david-voisin: It should be.
The status we display in databases-list is this one, as the do not get the inner Database info in the related query.

dbng/delete.go
sc-david-voisin: In both case we use progressive present, and both imply action in progress ?!

dbng/list.go
sc-david-voisin: The "owner" is the `currentUser`.

dbng/list.go
sc-david-voisin: I think so, can you develop this question ?

dbng/list.go
sc-david-voisin: No merit, for the messages I sticked to the specs as much as possible.
But good one indeed.

dbng/show.go
sc-david-voisin: This one provides more info, especially info that comes from the related `Database` entity, not only `App` or `AddOn` returned by `list`.
Especially, here yo can find version, force TLS, maintenance ... that are only available from Database entity.

cmd/databases_ng.go
sc-david-voisin: Extra remark, most of my `CurrentApp` are actually wrong, because the app is not specified.
I did not realize because I define a default one through `SCALINGO_APP` env var.

**But**, I prefer to leave it as it is, so as to produce conflicts/errors with Pierre's PR and identify quickly the calls to modify.

cmd/databases_ng.go
sc-david-voisin: same reply as above

cmd/databases_ng.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/Scalingo/cli/dbng"
+	"github.com/Scalingo/cli/detect"
+	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v8"
+)
+
+var (
+	databasesListCommand = cli.Command{
+		Name:     "databases",
+		Category: "Databases",
```
curzolapierre, line 18: IMO, this command could be in `Global` category.
```suggestion
		Category: "Global",
```

Note that this `Global` field is used by to generate the help.

cmd/databases_ng.go
```diff
@@ -0,0 +1,146 @@
...
+			return nil
+		},
+		ShellComplete: func(_ context.Context, c *cli.Command) {
+			_ = autocomplete.CmdFlagsAutoComplete(c, "databases")
+		},
+	}
+
+	databaseShowCommand = cli.Command{
+		Name:      "database-info",
+		Category:  "Databases",
```
curzolapierre, line 39: Here I really wonder if we'd prefer to create a dedicated category to make a clear distinction between database as addon and database 'NG'.

cmd/databases_ng.go
curzolapierre: We could indeed fix this in a dedicated PR once at least mine is merged (or both).

cmd/databases_ng.go
```diff
@@ -0,0 +1,146 @@
...
+			Description: "View database next generation detailed informations",
+			Examples:    []string{"scalingo database-info database_id"},
+			SeeAlso:     []string{"databases", "database-create", "database-destroy"},
+		}.Render(),
+		Action: func(ctx context.Context, c *cli.Command) error {
+			currentApp := detect.CurrentApp(c)
+			appID := c.Args().First()
+			if appID == "" {
+				return cli.ShowCommandHelp(ctx, c, "database-info")
+			}
```
curzolapierre, line 52: What about to be reflect what is done from `apps-info`?
i.e.
```sh
scalingo --app my-app apps-info
```
In that case this condition can be removed.

Or make the same change into `apps-info`.

cmd/databases_ng.go
```diff
@@ -0,0 +1,146 @@
...
+
+	databaseShowCommand = cli.Command{
+		Name:      "database-info",
+		Category:  "Databases",
+		Usage:     "View database next generation",
+		ArgsUsage: "database-id",
+		Description: CommandDescription{
+			Description: "View database next generation detailed informations",
+			Examples:    []string{"scalingo database-info database_id"},
+			SeeAlso:     []string{"databases", "database-create", "database-destroy"},
```
curzolapierre, line 45: We're using database name like for apps.

```suggestion
		ArgsUsage: "database",
		Description: CommandDescription{
			Description: "View database next generation detailed informations",
			Examples:    []string{"scalingo database-info database"},
			SeeAlso:     []string{"databases", "database-create", "database-destroy"},
```

cmd/databases_ng.go
curzolapierre: IMO this check should be removed. Neither an app or a database is existing at this step.

cmd/databases_ng.go
```diff
@@ -0,0 +1,146 @@
...
+		Description: CommandDescription{
+			Description: "Create a new database next generation",
+			Examples: []string{
+				"scalingo database-create --type postgresql --plan postgresql-dbng-starter-2048 my_super_database",
+				"scalingo database-create --type postgresql --plan postgresql-dbng-starter-2048 --wait my_super_database",
+			},
+			SeeAlso: []string{"databases", "database-info", "database-destroy"},
+		}.Render(),
+		Action: func(ctx context.Context, c *cli.Command) error {
+			currentApp := detect.CurrentApp(c)
```
curzolapierre, line 89: There is no app to detect here, this should be removed.

dbng/delete.go
curzolapierre: I'd be in favor of:
```sh
Database <db_name> has been deleted
```

If no error there is no doubt.

dbng/list.go
curzolapierre: At least this can be simplify: :sweat_smile:
```suggestion
	io.Warning("This command only displays databases next generation you own.")
```

cmd/databases_ng.go
curzolapierre: Otherwise I don't understand the usage of `currentApp` with `appID` from param. It must be one or the other never both.
e.g.
```sh
# the following makes non sense:
scalingo --database my-db database-info my-db
# either the same behavior as apps-info IIRC
scalingo --database my-db database-info
# or
scalingo database-info my-db
```

curzolapierre (commented):
Overall seems good to me, I've several questions.

Question: I see that you are using the term "database next generation". Shouldn't we just use "database"?

Nitpick: the branch name is not correct, it should mention the STORY-2512. And the notion automation is mentionning the STORY-2513.

cmd/databases_ng.go
sc-david-voisin: Can you explain this "Global" category ?
Going through the code I do not fully understand its usage, non to speak I've found mixed usages such as: "Addons - Global" or "Notifiers - Global" ?

cmd/databases_ng.go
sc-david-voisin: Why not, what about
"Database NG" or
"Database Next Generation" ?

cmd/databases_ng.go
sc-david-voisin: I disagree, the specs indicate the` Database ID` (in reality App ID):
https://www.notion.so/scalingooriginal/FEAT-CLI-List-and-create-and-update-database-261f536d89bf80a3a83ee588c1f94d89?source=copy_link#265f536d89bf80b1bc1bdca98ef61739

Also the `go-scalingo` matching methods:
https://www.notion.so/scalingooriginal/FEAT-SDK-Add-support-for-the-new-databases-related-actions-261f536d89bf809db0cff29aeebc22f4?source=copy_link#268f536d89bf8008af4ed4d4f7a1d315

cmd/databases_ng.go
leo-scalingo: That's a good question, eventually all databases will be adressable that way (with Database as First Class Citizen which should follow), I don't know if we want to make it generic now or not yet

cmd/databases_ng.go
leo-scalingo: I agree that behavior should be consistent between the 2 commands

cmd/databases_ng.go
leo-scalingo: :+1:

cmd/databases_ng.go
leo-scalingo: You need to ask for consent for commands which could allow to access user-data, when create a database, we're creating a resource for ourselves, there is no notion of user data.

cmd/databases_ng.go
sc-david-voisin: `currentApp` is wrong here, I use it for the `CheckForConsent` bellow.
It aims to become a `Resource` once merged with your PR.

dbng/add.go
leo-scalingo: I'm really not sure the App status will be updated, since the entity responsible for this is the `app-scheduler` and with DBNG it's never contacted, maybe there something I'm missing, but what is important is the status of the database only.

dbng/add.go
leo-scalingo: Not necessarily sure it's worth on my hand, actually if there is no timeout, just make sure Ctrl^C works correctly.

cmd/databases_ng.go
sc-david-voisin: I would avoid, app ID **is** the database ID, but must be have as a database ID.
The expected behavior is
```
scalingo database-info database_id
```

and ignore `--app`

cmd/databases_ng.go
sc-david-voisin: done

cmd/databases_ng.go
sc-david-voisin: same reply as above, used for the consent, to reworked during our PRs merge

cmd/databases_ng.go
sc-david-voisin: Ok to remove.

cmd/databases_ng.go
curzolapierre: But still, even if we talk about resource here, my example is still valid with `currentResource`.

If I reformulate: Otherwise I don't understand an usage of `currentResource` with a `databaseID` took param. It must be one or the other never both.

cmd/databases_ng.go
curzolapierre: `Global` is used to generate the help. Try:
```sh
scalingo help
```
At the bottom you can see Global commands. IMO `databases` should be in this category as well as `apps`.

dbng/add.go
sc-david-voisin: Ctrl^C works fine !

```
➜  ~ bsscalingo database-create --type postgresql --plan postgresql-dedicated-starter-1024 --wait test-dv-cli

-----> Your dedicated-pg database 68d5092645f8925a197c0048 ('test-dv-cli') is being provisioned…

⣽ Waiting for the database to be ready
^Cinterrupt catched, aborting…
➜  ~ 

```

dbng/add.go
sc-david-voisin: If we use `Database` status only, we may be disaligned with the output of `database-list` (`databases` command) that knows only  `App` entity (not  `Database`) and displays also the status.
Is it ok to cope with this disalignment ?

dbng/delete.go
sc-david-voisin: The addons use the same message, I can go for this one.

dbng/list.go
```diff
@@ -0,0 +1,55 @@
...
+	currentUser, err := config.C.CurrentUser(ctx)
+	if err != nil {
+		return errors.Wrap(ctx, err, "get current user")
+	}
+
+	for _, db := range databases {
+		role := utils.AppRole(currentUser, &db.App)
+
+		_ = t.Append([]string{
+			db.App.ID,
```
curzolapierre, line 40: It's should be the `UUID` here.

dbng/show.go
```diff
@@ -0,0 +1,53 @@
...
+		switch feature.Name {
+		case "force-ssl":
+			forceSSL = strings.ToLower(string(feature.Status))
+		case "publicly-available":
+			internetAccess = strings.ToLower(string(feature.Status))
+		}
+	}
+
+	data := [][]string{
+		{"ID", db.App.ID},
```
curzolapierre, line 38: Same as above, it should be UUID instead.

cmd/databases_ng.go
curzolapierre: Oh okay so other question: shouldn't be able to support the name **too** like what we have with apps?

```sh
scalingo apps-info my-app
scalingo database-info my-db
```

In addition, this would catch the apps-info on the hidden app.

cmd/databases_ng.go
curzolapierre: I'm not sure to understand the consent of what exactly?
Because database (or app) is not yet created so it could safely be removed IMHO.

cmd/databases_ng.go
matthieu526-scalingo: But it allows a nice message to the end user no? Otherwise it might remind you the usage, but without telling you what is wrong.

cmd/databases_ng.go
matthieu526-scalingo: So I guess you added "command", but left "databaseAddCommand" and not "databaseCreateCommand"?

cmd/databases_ng.go
matthieu526-scalingo: :information_source: This is exist. Not sure if you forgot or planned to do it later :slightly_smiling_face:

dbng/add.go
matthieu526-scalingo: Yes but this is pretty different still, so I wonder  if it still makes sense.

dbng/add.go
matthieu526-scalingo: Yes I know. But if the ID is also at the root, then this is more future proof. Meaning we can get rid of the `app` without changing the code as much.

dbng/add.go
matthieu526-scalingo: Wait, it raises another question then for the `list`. Does it work at the moment? Is the status properly reported into the `app`?

dbng/add.go
matthieu526-scalingo: Alright for, even though I don't see the value of this option :thinking: (actually I see it but I wonder  if people use it)

dbng/delete.go
matthieu526-scalingo: But aren't we doing that 6 hours later or something? Or this is only for mutualised one?

dbng/list.go
matthieu526-scalingo: Sorry David. Is the list returning DBs you own, or also the ones where you are a collaborator?

dbng/list.go
matthieu526-scalingo: Yes, but then the role is always "owner" so it can be hardcoded. There is no need to get the user. No?

dbng/show.go
matthieu526-scalingo: No I meant this exact line. The `Bulk` function.

dbng/add.go
curzolapierre: > I don't see the value of this option

Maybe from a CI pipeline, have a synchronous option is nice.

dbng/add.go
curzolapierre: AFAIK the app status should be always `new` because it cannot be deployed (in order to pass to `running`).

cmd/databases_ng.go
sc-david-voisin: My goal here is to check the user consent to access databases infos.
However the consent part has to reviewed once we merge both our PRs. I propose to leave comments were the consent seems appropriate, and address it in a third "merge" PR.

dbng/delete.go
sc-david-voisin: The deletion is applied straight away by the API.
So it appears deleted instantly on both CLI and Dashboard (just checked)... even if it takes hours to be really deleted: the real deletion is scheduled thereafter.

dbng/list.go
sc-david-voisin: I prefer to display a queried role. But you're right, as the user is always the owner maybe this column is simply useless ?

dbng/show.go
sc-david-voisin: `Bulk` is an other way to append rows in the table: all at once.

cmd/databases_ng.go
sc-david-voisin: Seems convincing, I apply :)

cmd/databases_ng.go
sc-david-voisin: I propose to distinguate the category by now, and to review later on (once we're not in Previw mode anymore?).

dbng/show.go
sc-david-voisin: ?

there is no UUID in App object, and we usually use ID for such column names !

PR #1148, opened 2025-09-23
Add command to set project ID for an application

matthieu526-scalingo:
This PR is based on [this](https://github.com/Scalingo/go-scalingo/pull/427) one and it should be merged before to  use the official release.


cmd/apps.go
```diff
@@ -58,4 +58,34 @@ var (
 			_ = autocomplete.CmdFlagsAutoComplete(c, "apps-info")
 		},
 	}
+
+	appsProjectSetCommand = cli.Command{
+		Name:      "project-set",
+		Category:  "App Management",
+		Usage:     "Set the project of an app",
+		ArgsUsage: "stack-id",
```
john-scalingo: Copy paste ;)

matthieu526-scalingo: Arf thanks :facepalm: . Done in d9a54e16541ebace892dc9dee0368b40bad00111.

apps/project.go
```diff
@@ -0,0 +1,25 @@
...
+	if err != nil {
+		return errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	_, err = c.AppsSetProject(ctx, appName, projectID)
+	if err != nil {
+		return errors.Wrap(ctx, err, "set project for app")
+	}
+
+	io.Statusf("Project ID has been set to %s on %s\n", projectID, appName)
```
sc-david-voisin:
```suggestion
	io.Statusf("Project ID has been set to %q for the application %s\n", projectID, appName)
```

sc-david-voisin: suggestion: I would be a little bit more explicit on this status message as per this proposal.

matthieu526-scalingo: I can do it, but it would introduce a "new behaviour" that doesn't exist on the CLI yet. So I think it is better to stay consistent. WDYT?

sc-david-voisin: What do you mean ? Specifying "application" ?
Here the "app" is specified in a status message in the same directory: https://github.com/Scalingo/cli/blob/master/apps/one_off_stop.go#L41

matthieu526-scalingo: Sorry, I focused on the `%q` to escape the characters. For "application" I'll add it.

sc-david-voisin: oh, Ok to keep `%s`, I did not understand

go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/Scalingo/cli
 
-go 1.24
+go 1.24.0
```
matthieu526-scalingo, line 3: This was brought by the latest version of `go-scalingo` because of an indirect dependency.. And I don't think I can get rid of it.

PR #1153, opened 2025-10-02
feat(database-ng) add support database flag to dbng commands

curzolapierre:

With --app and a database ID
```sh
➜ scalingo database-info --app $DATABASE_ID
 !     No database found. Please use --database flag.
```

With the correct --database flag
```sh
➜ scalingo database-info --database $DATABASE_ID
┌────────────────────────┬───────────────────────────────┐
│ ID                     │ 68c43e3cd9985668949c5ffa      │
...
```

With a database name
```sh
➜ scalingo apps-info --database $DATABASE
 !     It is currently impossible to use `apps-info` on a database.
```

The generated help
```sh
➜ scalingo database-info --help
NAME:
   scalingo database-info - View database next generation

USAGE:
   scalingo database-info database-id

CATEGORY:
   Databases NG

DESCRIPTION:
   View database next generation detailed informations

   Examples
     $ scalingo database-info database_id
     $ scalingo database-info --database database_id

   # See also 'databases' 'database-create' 'database-destroy'

OPTIONS:
   --database string  ID of the current database (default: "<database_name>")
   --region string    Name of the region to use
   --help, -h         show help
```

Also fixes STORY-2587

### DIFF
--- a/cmd/databases.go
+++ b/cmd/databases.go
@@ -23,7 +23,7 @@ var (
 		Category:  "Addons",
 		Usage:     "Enable a togglable feature from a database",
 		ArgsUsage: "feature-id",
-		Flags: []cli.Flag{&appFlag, &addonFlag, &cli.BoolFlag{
+		Flags: []cli.Flag{&appFlag, &addonFlag, databaseFlag(), &cli.BoolFlag{
 			Name:  "synchronous",
 			Usage: "Wait for the feature to be enabled synchronously",
 		}},
@@ -58,7 +58,7 @@ var (
 		Category:  "Addons",
 		Usage:     "Enable a togglable feature from a database",
 		ArgsUsage: "feature-id",
-		Flags:     []cli.Flag{&appFlag, &addonFlag},
+		Flags:     []cli.Flag{&appFlag, &addonFlag, databaseFlag()},
 		Description: CommandDescription{
 			Description: "Disable a togglable feature from a database",
 			Examples: []string{
@@ -88,7 +88,7 @@ var (
 		Name:     "backups-config",
 		Category: "Addons",
 		Usage:    "Configure the periodic backups of a database",
-		Flags: []cli.Flag{&appFlag, &addonFlag, &cli.StringFlag{
+		Flags: []cli.Flag{&appFlag, &addonFlag, databaseFlag(), &cli.StringFlag{
 			Name:  "schedule-at",
 			Usage: "Enable daily backups and schedule them at the specified hour of the day (in local time zone). It is also possible to specify the timezone to use.",
 		}, &cli.BoolFlag{
@@ -153,7 +153,7 @@ var (
 		Aliases:  []string{"database-list-users"},
 		Category: "Addons",
 		Usage:    "Print database's users",
-		Flags:    []cli.Flag{&appFlag, &addonFlag},
+		Flags:    []cli.Flag{&appFlag, &addonFlag, databaseFlag()},
 		Description: CommandDescription{
 			Description: `List the users of a database
 
@@ -183,7 +183,7 @@ Only available on ` + fmt.Sprintf("%s", dbUsers.SupportedAddons),
 		Category:  "Addons",
 		ArgsUsage: "user",
 		Usage:     "Delete a database's user",
-		Flags:     []cli.Flag{&appFlag, &addonFlag},
+		Flags:     []cli.Flag{&appFlag, &addonFlag, databaseFlag()},
 		Description: CommandDescription{
 			Description: `Delete the given user of a database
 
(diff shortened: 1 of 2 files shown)


utils/consent.go
```diff
@@ -84,7 +84,7 @@ func CheckForConsent(ctx context.Context, resourceName string, consentTypes ...C
 	}
 
 	for _, db := range dbs {
-		if db.App.Name == resourceName {
+		if db.App.Name == resourceName || db.App.ID == resourceName {
```
curzolapierre, line 87: Here both check are needed because:
- the database ID is used for all `scalingo database*` commands.
- the database name is used for all others (like what we do with `apps`).

cmd/databases_ng.go
```diff
@@ -39,18 +42,26 @@ var (
 		ArgsUsage: "database-id",
 		Description: CommandDescription{
 			Description: "View database next generation detailed informations",
-			Examples:    []string{"scalingo database-info database_id"},
-			SeeAlso:     []string{"databases", "database-create", "database-destroy"},
+			Examples: []string{
+				"scalingo database-info database_id",
+				"scalingo database-info --database database_id",
```
matthieu526-scalingo, line 49: non-blocking: I don't remember if this was in the specs, but when seeing this I wonder if this is a good idea to allow both options.

cmd/databases_ng.go
```diff
@@ -39,18 +42,26 @@ var (
...
+			},
+			SeeAlso: []string{"databases", "database-create", "database-destroy"},
 		}.Render(),
 		Action: func(ctx context.Context, c *cli.Command) error {
-			appID := c.Args().First()
-			if appID == "" {
+			databaseID := c.Args().First()
+			if databaseID == "" {
+				databaseID, _ = detect.GetCurrentDatabase(ctx, c)
```
matthieu526-scalingo, line 56: question: did you not check the error on purpose?

cmd/databases_ng.go
```diff
@@ -77,8 +88,8 @@ var (
 		Description: CommandDescription{
 			Description: "Create a new database next generation",
 			Examples: []string{
-				"scalingo database-create --type postgresql --plan postgresql-dbng-starter-2048 my_super_database",
-				"scalingo database-create --type postgresql --plan postgresql-dbng-starter-2048 --wait my_super_database",
+				"scalingo database-create --type postgresql-ng --plan postgresql-ng-enterprise-4096 my_super_database",
```
matthieu526-scalingo, line 93: question: I don't think the type is different, is it?
```suggestion
				"scalingo database-create --type postgresql --plan postgresql-ng-enterprise-4096 my_super_database",
```

detect/app.go
```diff
@@ -18,6 +18,20 @@ import (
...
 var errDatabaseNotFound = stderrors.New("database name not found")
 
+// GetCurrentDatabase is a helper to get the current database name and UUID.
+func GetCurrentDatabase(ctx context.Context, c *cli.Command) (string, string) {
+	currentDatabase, databaseUUID, err := currentDatabaseNameAndUUID(ctx, c)
+	if err != nil && !errors.Is(err, errDatabaseNotFound) {
+		io.Error(err)
+		os.Exit(1)
+	}
+	if err != nil {
```
matthieu526-scalingo, line 28: suggestion: I think it should be `errors.Is(err, errDatabaseNotFound)`. It is the only possible case left.

cmd/databases_ng.go
curzolapierre: As talked during last PSM, we want to use this PR as support for this topic precisely. So your comment is welcome 👍 

To answer, IMO we could keep both.
The usage of flags (i.e. `--database` here) was not in the specs, but they are easier to use and that's currently how we do for other commands.

e.g. equivalent command for apps:

```sh
scalingo apps-info --app <app_name>
```

If we had to choose one, I would go with the use of flags (`--database`). But I'll let @leo-scalingo give his opinion.

cmd/databases_ng.go
curzolapierre: `GetCurrentDatabase` will exit on error. The second param here is the database UUID.

cmd/databases_ng.go
curzolapierre: It is since the first deployment of given plans. We now distinguish between the two.
Note: there were a conflict of credentials (cf: STORY-2622)

detect/app.go
curzolapierre: Right,

Done here: cbd0da0283f331ba9bf257e532f2305e08857fe1

cmd/databases_ng.go
leo-scalingo: I also prefer using flags than arguments, for this precise command, since the database_id is the subject I'm ok to keep both as syntax sugar

PR #1154, opened 2025-10-06
fix(privatenetworks): default format should be `table`

EtienneM:
`table` is the default format in all the CLI. I think it should still be the same.

### DIFF
--- a/cmd/privatenetworks.go
+++ b/cmd/privatenetworks.go
@@ -26,7 +26,7 @@ var (
 		Flags: []cli.Flag{&appFlag,
 			&cli.StringFlag{
 				Name:  "format",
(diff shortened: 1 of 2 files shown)


cmd/privatenetworks.go
```diff
@@ -26,7 +26,7 @@ var (
 		Flags: []cli.Flag{&appFlag,
 			&cli.StringFlag{
 				Name:  "format",
-				Value: "json",
+				Value: outputFormatTable,
```
sc-zenokerr, line 29: Thanks for using one of the consts here.

sc-zenokerr (approved):
LGTM

PR #1171, opened 2025-12-31
[STORY-3093] feat(dbng): add firewall rules commands

briceamen:
## Summary
Add firewall rules management commands for DBNG databases:
- `database-firewall-rules` - List firewall rules
- `database-firewall-rules-add` - Add custom CIDR or managed range rule
- `database-firewall-rules-remove` - Remove a rule
- `database-firewall-managed-ranges` - List available managed ranges

Commands support both positional arguments and `--database` flag, matching existing DBNG command patterns (`database-info`, `database-destroy`).

### Usage examples
```bash
# Positional argument style
scalingo database-firewall-rules my-db-id
scalingo database-firewall-rules-add my-db-id --cidr 203.0.113.0/24 --label "Office"
scalingo database-firewall-rules-remove my-db-id rule-id

# --database flag style
scalingo --database my-db database-firewall-rules
scalingo --database my-db database-firewall-rules-add --managed-range man-osc-fr1-egress
scalingo --database my-db database-firewall-rules-remove rule-id
```

### Changes
- Added `GetAddonIDFromDatabase` helper in `detect/app.go` to resolve addon ID from database ID
- Firewall commands now support positional args like other DBNG commands

## Test plan
Tested on staging (`osc-st-fr1`) with database `test-firewall-rules-db`:

**Positional argument style:**
- [x] `database-firewall-rules test-firewall-rules-db` - List rules
- [x] `database-firewall-managed-ranges test-firewall-rules-db` - List managed ranges
- [x] `database-firewall-rules-add test-firewall-rules-db --cidr 203.0.113.0/24 --label "Test office"` - Add custom rule
- [x] `database-firewall-rules-remove test-firewall-rules-db <rule-id>` - Remove rule

**--database flag style:**
- [x] `--database test-firewall-rules-db database-firewall-rules` - List rules
- [x] `--database test-firewall-rules-db database-firewall-rules-add --managed-range man-osc-st-fr1-egress` - Add managed range
- [x] `--database test-firewall-rules-db database-firewall-rules-remove <rule-id>` - Remove rule

### DIFF
--- a/cmd/firewall_rules.go
+++ b/cmd/firewall_rules.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 
 	"github.com/urfave/cli/v3"
 
@@ -30,11 +31,11 @@ var (
 		}.Render(),
 		Action: func(ctx context.Context, c *cli.Command) error {
 			databaseName, addonID, err := detect.GetDatabaseFromArgs(ctx, c)
+			if errors.Is(err, detect.ErrTooManyArguments) {
+				io.Error(err)
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules")
+			}
 			if err != nil {
-				if err == detect.ErrTooManyArguments {
-					io.Error(err)
-					return cli.ShowCommandHelp(ctx, c, "database-firewall-rules")
-				}
 				errorQuit(ctx, err)
 			}
 
@@ -74,11 +75,11 @@ var (
 		}.Render(),
 		Action: func(ctx context.Context, c *cli.Command) error {
 			databaseName, addonID, err := detect.GetDatabaseFromArgs(ctx, c)
+			if errors.Is(err, detect.ErrTooManyArguments) {
+				io.Error(err)
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
+			}
 			if err != nil {
-				if err == detect.ErrTooManyArguments {
-					io.Error(err)
-					return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
-				}
 				errorQuit(ctx, err)
 			}
 
@@ -96,10 +97,6 @@ var (
 				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
 			}
 
-			if managedRange != "" && label != "" {
-				io.Warning("--label flag is ignored for managed ranges")
-			}
-
 			utils.CheckForConsent(ctx, databaseName, utils.ConsentTypeDBs)
 
 			var params scalingo.FirewallRuleCreateParams
@@ -113,6 +110,7 @@ var (
 				params = scalingo.FirewallRuleCreateParams{
 					Type:    scalingo.FirewallRuleTypeManagedRange,
 					RangeID: managedRange,
+					Label:   label,
 				}
 			}
 
@@ -206,11 +204,11 @@ var (
 		}.Render(),
 		Action: func(ctx context.Context, c *cli.Command) error {
 			databaseName, addonID, err := detect.GetDatabaseFromArgs(ctx, c)
+			if errors.Is(err, detect.ErrTooManyArguments) {
+				io.Error(err)
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-managed-ranges")
+			}
 			if err != nil {
-				if err == detect.ErrTooManyArguments {
-					io.Error(err)
-					return cli.ShowCommandHelp(ctx, c, "database-firewall-managed-ranges")
-				}
 				errorQuit(ctx, err)
 			}
 


detect/app.go
```diff
@@ -18,6 +18,31 @@ import (
...
 
+// GetAddonIDFromDatabase resolves the addon ID from a database ID by calling the API.
+// This is useful when the database ID is provided as a positional argument.
+func GetAddonIDFromDatabase(ctx context.Context, databaseID string) (string, error) {
+	client, err := config.ScalingoClient(ctx)
+	if err != nil {
+		return "", errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	// In case of a database, addons list should return only one element.
```
sc-david-voisin: question: why only one element implies a database ? Even more curious, when can there be multiple addons for one database ID ?

dbng/firewall_rules.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"context"
+	"os"
+
+	"github.com/olekukonko/tablewriter"
+
+	"github.com/Scalingo/cli/config"
+	"github.com/Scalingo/cli/io"
+	"github.com/Scalingo/go-scalingo/v8"
+	"github.com/Scalingo/go-utils/errors/v2"
```
sc-david-voisin, line 12: suggestion: what about introducing `errors/v3` here ?

dbng/firewall_rules.go
```diff
@@ -0,0 +1,114 @@
...
+		return errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	rule, err := c.Preview().FirewallRulesCreate(ctx, databaseID, addonID, params)
+	if err != nil {
+		return errors.Wrap(ctx, err, "add firewall rule")
+	}
+
+	io.Statusf("Firewall rule '%s' has been added to database '%s'.\n", rule.ID, databaseID)
+	io.Warning("Firewall rules take time to be applied due to infrastructure provisioning.")
```
sc-david-voisin: Maybe reformulate a bit. A proposition:

"Expect some delay for the firewall rules to be applied, due to infrastructure provisioning."

detect/app.go
briceamen: The `AddonsList` API works for both apps and DBNG databases (same endpoint `/apps/{id}/addons`).

- When called with a DBNG database ID: returns exactly 1 addon (the database itself is modeled as an app with a single addon)
- When called with an application ID: returns all addons attached to the app (postgresql, redis, etc.) - potentially multiple

So `len(addons) > 1` is a safeguard to detect if someone accidentally passed an app ID instead of a database ID.

This is a bit hacky I know, I'll make the comment clearer to explain this distinction.

dbng/firewall_rules.go
briceamen: We could but switching would require updating all dbng files for consistency

sc-david-voisin: Ok, for a next time !

detect/app.go
sc-david-voisin: Thanks for the clear explanation and the comment :)

cmd/firewall_rules.go
```diff
@@ -0,0 +1,257 @@
...
+				addonID, err = detect.GetAddonIDFromDatabase(ctx, databaseID)
+				if err != nil {
+					errorQuit(ctx, err)
+				}
+			default:
+				io.Error("Too many arguments")
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules")
+			}
+
+			utils.CheckForConsent(ctx, databaseID, utils.ConsentTypeDBs)
```
matthieu526-scalingo: question: is the consent required here?

cmd/firewall_rules.go
```diff
@@ -0,0 +1,257 @@
...
+				// database-id provided as positional arg
+				databaseID = c.Args().First()
+				addonID, err = detect.GetAddonIDFromDatabase(ctx, databaseID)
+				if err != nil {
+					errorQuit(ctx, err)
+				}
+			default:
+				io.Error("Too many arguments")
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
+			}
```
matthieu526-scalingo: question: this is done 3 times? What do you about about making a small function out it?

cmd/firewall_rules.go
```diff
@@ -0,0 +1,257 @@
...
+			if cidr != "" {
+				params = scalingo.FirewallRuleCreateParams{
+					Type:  scalingo.FirewallRuleTypeCustomRange,
+					CIDR:  cidr,
+					Label: label,
+				}
+			} else {
+				params = scalingo.FirewallRuleCreateParams{
+					Type:    scalingo.FirewallRuleTypeManagedRange,
+					RangeID: managedRange,
```
matthieu526-scalingo, line 112: question: here the label isn't used, is it on purpose? Otherwise, shouldn't we block the flag for manages rules?

cmd/firewall_rules.go
```diff
@@ -0,0 +1,257 @@
...
+					errorQuit(ctx, err)
+				}
+			default:
+				io.Error("Too many arguments")
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-managed-ranges")
+			}
+
+			utils.CheckForConsent(ctx, databaseID, utils.ConsentTypeDBs)
+
+			err = dbng.FirewallManagedRangesList(ctx, databaseID, addonID)
```
matthieu526-scalingo: question: why do you need the DB ID to list the ranges?

dbng/firewall_rules.go
```diff
@@ -0,0 +1,114 @@
...
+		return errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	rule, err := c.Preview().FirewallRulesCreate(ctx, databaseID, addonID, params)
+	if err != nil {
+		return errors.Wrap(ctx, err, "add firewall rule")
+	}
+
+	io.Statusf("Firewall rule '%s' has been added to database '%s'.\n", rule.ID, databaseID)
+	io.Warning("Expect some delay for the firewall rules to be applied, due to infrastructure provisioning.")
```
matthieu526-scalingo: question: is this the good wording? We are not really provisioning anything at this point/

dbng/firewall_rules.go
```diff
@@ -0,0 +1,114 @@
...
+		return errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	err = c.Preview().FirewallRulesDestroy(ctx, databaseID, addonID, ruleID)
+	if err != nil {
+		return errors.Wrap(ctx, err, "remove firewall rule")
+	}
+
+	io.Statusf("Firewall rule '%s' has been removed from database '%s'.\n", ruleID, databaseID)
+	io.Warning("Expect some delay for the firewall rules to be applied, due to infrastructure provisioning.")
```
matthieu526-scalingo: nitpick: do we want to use `applied` here?

cmd/firewall_rules.go
curzolapierre: IMO yes. But why not?

cmd/firewall_rules.go
curzolapierre: It is needed to generate the token to contact related API.

cmd/firewall_rules.go
```diff
@@ -0,0 +1,257 @@
...
+
+			switch c.Args().Len() {
+			case 0:
+				// No positional arg, database from --database flag
+				databaseID, addonID = detect.GetCurrentDatabase(ctx, c)
+			case 1:
+				// database-id provided as positional arg
+				databaseID = c.Args().First()
+				addonID, err = detect.GetAddonIDFromDatabase(ctx, databaseID)
+				if err != nil {
```
curzolapierre: Semantically we used database "name" instead of ID. Could you apply the change elsewhere in this file?

Or do I miss something?

```suggestion
			var databaseName, addonID string
			var err error

			switch c.Args().Len() {
			case 0:
				// No positional arg, database from --database flag
				databaseName, addonID = detect.GetCurrentDatabase(ctx, c)
			case 1:
				// database-name provided as positional arg
				databaseName = c.Args().First()
				addonID, err = detect.GetAddonIDFromDatabase(ctx, databaseName)
				if err != nil {
```

detect/app.go
```diff
@@ -18,6 +18,34 @@ import (
...
+	client, err := config.ScalingoClient(ctx)
+	if err != nil {
+		return "", errors.Wrap(ctx, err, "get Scalingo client")
+	}
+
+	// AddonsList works for both apps and DBNG databases (same API endpoint).
+	// A DBNG database is modeled as an app with a single addon (itself),
+	// whereas an application can have multiple addons (postgresql, redis, etc.).
+	// If multiple addons are returned, the ID is likely an application, not a database.
+	addons, err := client.AddonsList(ctx, databaseID)
```
curzolapierre: Like before, this should be the `databaseName`

```suggestion
	addons, err := client.AddonsList(ctx, databaseName)
```

cmd/firewall_rules.go
matthieu526-scalingo: Because I am not sure how the consent should be used in that case. This is for a database setting, not accessing the data?

cmd/firewall_rules.go
```diff
@@ -0,0 +1,231 @@
...
+		Action: func(ctx context.Context, c *cli.Command) error {
+			databaseName, addonID, err := detect.GetDatabaseFromArgs(ctx, c)
+			if err != nil {
+				if err == detect.ErrTooManyArguments {
+					io.Error(err)
+					return cli.ShowCommandHelp(ctx, c, "database-firewall-rules")
+				}
+				errorQuit(ctx, err)
+			}
+
```
curzolapierre, line 41: Suggestion: Could be simplified by:
```suggestion
			if errors.Is(err, detect.ErrTooManyArguments) {
				io.Error(err)
				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules")
			}
			if err != nil {
				errorQuit(ctx, err)
			}

```
In addition, this is a good practice to use `errors.Is` instead of direct comparison.

cmd/firewall_rules.go
```diff
@@ -0,0 +1,231 @@
...
+			if cidr != "" && managedRange != "" {
+				io.Error("Cannot specify both --cidr and --managed-range")
+				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
+			}
+
+			if managedRange != "" && label != "" {
+				io.Warning("--label flag is ignored for managed ranges")
+			}
```
curzolapierre: **Question**: are you sure we cannot add label to managed ranges?

cmd/firewall_rules.go
```diff
@@ -221,28 +205,18 @@ var (
...
+			if err != nil {
+				if err == detect.ErrTooManyArguments {
+					io.Error(err)
+					return cli.ShowCommandHelp(ctx, c, "database-firewall-managed-ranges")
 				}
-			default:
-				io.Error("Too many arguments")
-				return cli.ShowCommandHelp(ctx, c, "database-firewall-managed-ranges")
+				errorQuit(ctx, err)
 			}
```
curzolapierre, line 213: Same as above

cmd/firewall_rules.go
```diff
@@ -83,23 +73,13 @@ var (
...
+			databaseName, addonID, err := detect.GetDatabaseFromArgs(ctx, c)
+			if err != nil {
+				if err == detect.ErrTooManyArguments {
+					io.Error(err)
+					return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
 				}
-			default:
-				io.Error("Too many arguments")
-				return cli.ShowCommandHelp(ctx, c, "database-firewall-rules-add")
+				errorQuit(ctx, err)
```
curzolapierre, line 83: Same as above

cmd/firewall_rules.go
briceamen: Oups you're right `FirewallRuleCreateParams` has Label as a general field (not specific to custom ranges), and `FirewallRule` response also has Label for all types. Will change

cmd/firewall_rules.go
briceamen: Much cleaner indeed. Thanks.

dbng/firewall_rules.go
matthieu526-scalingo: @briceamen up here. Is applied applicable for removal?

curzolapierre (approved):
Thanks! LGTM

PR #1178, opened 2026-01-13
[STORY-3236] Add DBNG plan upgrade command

briceamen:
## Summary
- add database NG plan upgrade command
- implement plan change wait logic and upgrade flow
- update command registration and changelog

## Testing
- manual: created dbng and upgraded plan on staging

### DIFF
--- a/cmd/databases_ng.go
+++ b/cmd/databases_ng.go
@@ -143,8 +143,10 @@ var (
 			var plan string
 
 			switch len(args) {
+			case 0:
+				return cli.ShowCommandHelp(ctx, c, "database-upgrade")
 			case 1:
-				if c.IsSet("database") || os.Getenv("SCALINGO_DATABASE") != "" {
+				if databaseNameFromFlags(c) != "" || os.Getenv("SCALINGO_DATABASE") != "" {
 					databaseID, _ = detect.GetCurrentDatabase(ctx, c)
 					plan = args[0]
 				} else {
--- a/dbng/upgrade.go
+++ b/dbng/upgrade.go
@@ -24,6 +24,11 @@ func Upgrade(ctx context.Context, databaseID, plan string, wait bool) error {
 		return errors.Wrap(ctx, err, "show database")
 	}
 
+	if plan == db.Plan {
+		io.Statusf("Database %s ('%s') is already on plan %s.\n", db.ID, db.Name, db.Plan)
+		return nil
+	}
+
 	planID, err := utils.FindPlan(ctx, c, db.Technology, plan)
 	if err != nil {
 		return errors.Wrap(ctx, err, "invalid database plan")
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -37,6 +37,16 @@ func databaseFlag() *cli.StringFlag {
 	return &cli.StringFlag{}
 }
 
+func databaseNameFromFlags(c *cli.Command) string {
+	for _, cliContext := range c.Lineage() {
+		dbName := cliContext.String("database")
+		if dbName != "" && dbName != "<database_name>" {
+			return dbName
+		}
+	}
+	return ""
+}
+
 // addonUUIDFromFlags returns the addon UUID based on the --addon flag, the SCALINGO_ADDON env var
 // or the current detected database.
 // exitIfMissing is optional. Set to true to show a message requesting for the --addon flag.


dbng/upgrade.go
```diff
@@ -0,0 +1,119 @@
...
+	var dbAddon *scalingo.Addon
+	if db.Database.ResourceID != "" {
+		for _, addon := range addons {
+			if addon.ResourceID == db.Database.ResourceID {
+				dbAddon = addon
+				break
+			}
+		}
+	}
+	if dbAddon == nil && len(addons) == 1 {
```
matthieu526-scalingo: question: is this required? Shouldn't this be covered in the loop?

dbng/upgrade.go
```diff
@@ -0,0 +1,119 @@
...
+		db.Name,
+	)
+
+	if wait {
+		spin := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
+		spin.Suffix = " Waiting for the plan change to complete\n"
+		spin.Start()
+
+		err = waitForDatabasePlanChange(ctx, c, db.ID)
+		spin.Stop()
```
matthieu526-scalingo: question: shouldn't this be defered?

cmd/databases_ng.go
```diff
@@ -119,6 +120,69 @@ var (
...
+				} else {
+					io.Error("Missing required argument: plan")
+					return cli.ShowCommandHelp(ctx, c, "database-upgrade")
+				}
+			case 2:
+				databaseID = args[0]
+				plan = args[1]
+			default:
+				io.Error("Invalid number of arguments")
+				return cli.ShowCommandHelp(ctx, c, "database-upgrade")
```
curzolapierre, line 161: Nitpick: generally with no argument given we just print the help WDYT?

Otherwise, could you share an example of the output either here in comment or in PR's body?

dbng/upgrade.go
briceamen: Yep my bad Copilot made a good suggestion and my refactoring was half baked. Will fix

cmd/databases_ng.go
```diff
@@ -119,6 +120,69 @@ var (
...
+
+			switch len(args) {
+			case 1:
+				if c.IsSet("database") || os.Getenv("SCALINGO_DATABASE") != "" {
+					databaseID, _ = detect.GetCurrentDatabase(ctx, c)
+					plan = args[0]
+				} else {
+					io.Error("Missing required argument: plan")
+					return cli.ShowCommandHelp(ctx, c, "database-upgrade")
+				}
```
curzolapierre: **Question**: is this still working when we pass `--database` as first argument?

IIRC is it what is usually expected for application flag:
```sh
scalingo --database MY-DB-ID database-upgrade postgresql-ng-enterprise-8192
```

dbng/upgrade.go
```diff
@@ -0,0 +1,118 @@
...
+	return nil
+}
+
+func waitForDatabasePlanChange(ctx context.Context, client *scalingo.Client, databaseID string) error {
+	db, err := client.Preview().DatabaseShow(ctx, databaseID)
+	if err != nil {
+		return err
+	}
+
+	if db.Database.Status == scalingo.DatabaseStatusRunning {
```
curzolapierre, line 96: **Suggestion** I'd also check the plan. IIRC API still accept the request without error even if the requested plan is the same as the current one.
So in that specific case it'd hang forever.

dbng/upgrade.go
```diff
@@ -38,17 +38,16 @@ func Upgrade(ctx context.Context, databaseID, plan string, wait bool) error {
 	}
 
 	var dbAddon *scalingo.Addon
-	if db.Database.ResourceID != "" {
+	if len(addons) == 1 {
```
matthieu526-scalingo, line 46: nitpick: I am not sure this optimization is needed. The loop is probably not expensive.

matthieu526-scalingo (approved):

cmd/databases_ng.go
briceamen: Agree, it should behave like `database-create` and just print the help when no args are provided. I’ll align with that.

For the output, here’s a sample from staging:

```
$ scalingo database-upgrade 69652796697f13aac6687e51 postgresql-ng-enterprise-8192
Your postgresql-ng database 69652796697f13aac6687e51 ('ng-plan-change-xxx') is being upgraded…

```

With `--wait`, it shows the spinner then:
`Your database plan change is complete.`”

cmd/databases_ng.go
briceamen: Yes that’s the expected usage:
`scalingo --database MY-DB-ID database-upgrade postgresql-ng-enterprise-8192`
It work since the flag is global.

cmd/databases_ng.go
curzolapierre: You didn't address my first comment, just to be sure if it was intended because you mentioned that wou'll align with `database-create` command.

PR #1199, opened 2026-03-06
feat: add JSON output for the `apps` command

### LINKED ISSUE #512: Improve scriptability of the tool by proposing JSON output
It would prevent user to have to base their scripts on some table display grep/awk/etc clumsy scripts.

EtienneM:
For now I just want to introduce this pattern in a single command in order to have inputs on the code organization. The purpose would be that after this PR, it's easy to slowly update the existing commands.

Fix #512

### DIFF
--- a/apps/list.go
+++ b/apps/list.go
@@ -2,18 +2,17 @@ package apps
 
 import (
 	"context"
-	"fmt"
-	"os"
-
-	"github.com/olekukonko/tablewriter"
 
 	"github.com/Scalingo/cli/config"
-	"github.com/Scalingo/cli/io"
-	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v10"
 	"github.com/Scalingo/go-utils/errors/v3"
 )
 
-func List(ctx context.Context, projectSlug string) error {
+type ListRenderer interface {
+	Render(ctx context.Context, apps []*scalingo.App) error
+}
+
+func List(ctx context.Context, renderer ListRenderer, projectSlug string) error {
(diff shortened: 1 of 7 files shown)


apps/list.go
```diff
@@ -2,18 +2,17 @@ package apps
...
+	"github.com/Scalingo/go-scalingo/v10"
 	"github.com/Scalingo/go-utils/errors/v3"
 )
 
-func List(ctx context.Context, projectSlug string) error {
+type ListRenderer interface {
+	Render(ctx context.Context, apps []*scalingo.App) error
+}
+
+func List(ctx context.Context, renderer ListRenderer, projectSlug string) error {
```
EtienneM, line 15: What I like with this pattern is that the `List` command does not know anything about the format. It just need to calls `Render`.

apps/list.go
```diff
@@ -2,18 +2,17 @@ package apps
...
 
 	"github.com/Scalingo/cli/config"
-	"github.com/Scalingo/cli/io"
-	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v10"
 	"github.com/Scalingo/go-utils/errors/v3"
 )
 
-func List(ctx context.Context, projectSlug string) error {
+type ListRenderer interface {
```
EtienneM, line 11: What I don't like is that it would require an interface per command. I don't really like that. I would love to have a single interface whatever the command. And the command only calls `Render(ctx)`. Then we would need a way to inject the data (list of apps or anything else depending on the command) in the renderer.

EtienneM:
Replaced by https://github.com/Scalingo/cli/pull/1204

PR #1204, opened 2026-03-11
feat: add JSON output for the `apps` command

EtienneM:
This is another implementation compared to the one in #1199. In this one I introduce a generic interface `Renderer`. The next step would be to update all `cmd/*` functions to take a renderer as argument and move the rendering in the `internal/boundaries/out/renderer/table` package.

Related to #512

### DIFF
--- a/internal/boundaries/out/renderer/table/apps_list.go
+++ b/internal/boundaries/out/renderer/table/apps_list.go
@@ -1,0 +1,50 @@
+package table
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/olekukonko/tablewriter"
+
+	"github.com/Scalingo/cli/internal/boundaries/out/renderer"
+	"github.com/Scalingo/cli/io"
+	"github.com/Scalingo/cli/utils"
+	"github.com/Scalingo/go-scalingo/v11"
+	"github.com/Scalingo/go-utils/errors/v3"
+)
+
+type appsListRenderer struct {
+	currentUser *scalingo.User
+	apps        []*scalingo.App
+}
+
+func NewAppsList(currentUser *scalingo.User) renderer.Renderer[[]*scalingo.App] {
+	return &appsListRenderer{
+		currentUser: currentUser,
+	}
+}
+
+func (r *appsListRenderer) Render(ctx context.Context) error {
+	if len(r.apps) == 0 {
+		fmt.Println(io.Indent("\nYou haven't created any app yet, create your first application using:\n→ scalingo create <app_name>\n", 2))
+		return nil
+	}
+
+	t := tablewriter.NewWriter(os.Stdout)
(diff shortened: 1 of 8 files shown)


internal/boundaries/out/renderer/table/apps_list.go
```diff
@@ -0,0 +1,50 @@
...
+
+func NewAppsList(currentUser *scalingo.User) renderer.Renderer[[]*scalingo.App] {
+	return &appsListRenderer{
+		currentUser: currentUser,
+	}
+}
+
+func (r *appsListRenderer) Render(ctx context.Context) error {
+	if len(r.apps) == 0 {
+		fmt.Println(io.Indent("\nYou haven't created any app yet, create your first application using:\n→ scalingo create <app_name>\n", 2))
```
sc-zenokerr: Question: When filtering apps by project, is this message still valid?

Is it possible the user may have created apps that are not in the project?

scalingo/main.go
```diff
@@ -88,6 +89,7 @@ func main() {
 	app.Flags = []cli.Flag{
 		&cli.StringFlag{Name: "addon", Value: "<addon_id>", Usage: "ID of the current addon", Sources: cli.EnvVars("SCALINGO_ADDON")},
 		&cli.StringFlag{Name: "app", Aliases: []string{"a"}, Value: "<name>", Usage: "Name of the app", Sources: cli.EnvVars("SCALINGO_APP")},
+		&cli.StringFlag{Name: "format", Value: string(renderer.FormatTable), Usage: "[" + string(renderer.FormatJSON) + "|" + string(renderer.FormatTable) + "]"},
```
sc-zenokerr, line 92: Question:  will this root `--format` flag conflict with other exiting `--format` flags?

sc-zenokerr (changes requested):
A couple of questions

scalingo/main.go
EtienneM: Good question! I double checked and I confirm that the `format` command declared for the `private-networks-domain-names` command keeps working

internal/boundaries/out/renderer/table/apps_list.go
EtienneM: mmmh that's a good point 🤔 

Would the following phrasing be better for you?
```suggestion
		fmt.Println(io.Indent("\nYou haven't created any app yet in this project, create your first application using:\n→ scalingo create <app_name>\n", 2))
```

Or you thought about something else?

sc-zenokerr: That works.

scalingo/main.go
sc-zenokerr: Thanks nice.

sc-zenokerr (approved):
LGTM

PR #1206, opened 2026-03-12
fix(databases): `parseScheduleAtFlag` returns a validation error

### LINKED ISSUE #1106: backups-config: the "schedule-at" flag is not properly parsed
When using the `scalingo` CLI, we document it's possible to configure Scheduled backup by running this command:
```bash
scalingo --app my-app --addon <whatever> backups-config --schedule-at "4:00 UTC"
```

So the `schedule-at` flag is supposed to support passing a time + timezone.

In facts, when doing so, the CLI reports the following error:
```text
 !     An error occurred:
       fail to parse the schedule-at flag
```

I see 2 issues:
- The error message should report "fail to parse the schedule-at **value**" because the flag is indeed OK.
- We should either:
  - actually support passing a time + timezone
  - or remove this possibility from the CLI help and from our documentation.

For the record, this command works:
```bash
scalingo --app my-app --addon <whatever> backups-config --schedule-at 4
```

But it's also badly documented.

EtienneM:
The error message is now clearer if the client does not provide a good format. For example:

```
 bsscalingo --app biniou backups-config --addon postgresql --schedule-at "azerty UTC"
 !     An error occurred:
       schedule-at=first field must be a digit representing the hour
```

And I updated the parsing function to be compatible with `4:00` (and not only `4`).

Fix #1106 

### DIFF
--- a/cmd/databases.go
+++ b/cmd/databases.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -15,6 +14,7 @@ import (
 	"github.com/Scalingo/cli/detect"
 	"github.com/Scalingo/cli/utils"
 	"github.com/Scalingo/go-scalingo/v10"
+	"github.com/Scalingo/go-utils/errors/v3"
 )
 
 var (
@@ -42,7 +42,7 @@ var (
 
 			addonName := addonUUIDFromFlags(ctx, c, currentResource, true)
 			if c.NArg() != 1 {
-				errorQuit(ctx, errors.New("feature argument should be specified"))
+				errorQuit(ctx, errors.New(ctx, "feature argument should be specified"))
 			}
 			feature := c.Args().First()
 			err := db.EnableFeature(ctx, c, currentResource, addonName, feature)
@@ -73,7 +73,7 @@ var (
 
 			addonName := addonUUIDFromFlags(ctx, c, currentResource, true)
 			if c.NArg() != 1 {
-				errorQuit(ctx, errors.New("feature argument should be specified"))
+				errorQuit(ctx, errors.New(ctx, "feature argument should be specified"))
 			}
 			feature := c.Args().First()
 			err := db.DisableFeature(ctx, currentResource, addonName, feature)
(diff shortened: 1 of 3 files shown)


cmd/databases.go
```diff
@@ -296,26 +296,39 @@ Only available on ` + fmt.Sprintf("%s", dbUsers.SupportedAddons),
...
-		return -1, nil, errors.New("parse the schedule-at value")
+		validationError = validationError.Set("schedule-at", "only contains two space-separated fields")
+		return -1, nil, validationError.Build()
 	}
-	scheduleAt, err = strconv.Atoi(s[0])
+	scheduleAtTime := s[0]
+	scheduleAtTimezone := s[1]
+
+	// If client writes "4:00", we only want to keep "4" in scheduleAtTime
+	scheduleAtTimeSplit := strings.SplitN(scheduleAtTime, ":", 2)
```
sc-zenokerr: Question, do we want to be prescriptive or accommodating about the format allowed?

If we allow `4` and `4:00` do we want to allow `04` and `04:00` or `4h` and `4h00` etc... where is the line drawn.

If we want to be accommodating I would suggest replacing the `scheduleAtTimeSplit` with something more general, such as: 

```
re := regexp.MustCompile(`([:hH].*)|[0]+`)
scheduleAtHour, err := strconv.Atoi(re.ReplaceAllString(scheduleAtTime, ""))
``` 

as this would accommodate lots of different formats e.g:
```
		"4",
		"4h",
		"4H",
		"04h",
		"4:00",
		"4:12",
		"04:00",
		"04:12",
		"09:00",
		"09",
		"4h:45",
```

I made a goplayground here to test this for myself: https://go.dev/play/p/UxHf9xVmiCJ

sc-zenokerr (changes requested):
A suggestion otherwise LGTM.

cmd/databases.go
EtienneM: Great idea, thanks. Done in 40bdb7db. I also took the opportunity to add a unit test for `parseScheduleAtFlag` to ensure we don't break anything.

sc-zenokerr (approved):
LGTM. Thanks :)

PR #1211, opened 2026-04-08
fix: avoid panic when polling one-off operations

### LINKED ISSUE #1207: Panic when starting a one-off container
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x9b178d]

goroutine 1 [running]:
github.com/Scalingo/cli/apps.(*OperationWaiter).WaitOperation(0xc00026a7c0, {0xd42c48, 0xc0003f3620})
        /home/runner/work/cli/cli/apps/operations.go:100 +0xb4d
github.com/Scalingo/cli/apps.Run({0xd42c48, 0xc0003f3620}, {{0x7ffe2d7258d1, 0x8}, {0x0, 0x0}, 0x0, 0x0, {0x0, 0x0}, ...})
        /home/runner/work/cli/cli/apps/run.go:144 +0x906
github.com/Scalingo/cli/cmd.init.func221({0xd42c48, 0xc0003f3620}, 0x128b3e0)
        /home/runner/work/cli/cli/cmd/run.go:117 +0x4c8
github.com/Scalingo/cli/cmd.NewAppCommands.(*AppCommands).addCommand.regionalCommandAction.func1({0xd42c48, 0xc0003f3620}, 0x128b3e0)
        /home/runner/work/cli/cli/cmd/commands.go:116 +0x377
github.com/urfave/cli/v3.(*Command).run(0x128b3e0, {0xd42c48, 0xc0003f3620}, {0xc0003f3560, 0x3, 0x3})
        /home/runner/work/cli/cli/vendor/github.com/urfave/cli/v3/command_run.go:360 +0x2b6a
github.com/urfave/cli/v3.(*Command).run(0xc0001f4b08, {0xd42c48, 0xc0003d5590}, {0xc000032100, 0x8, 0x8})
        /home/runner/work/cli/cli/vendor/github.com/urfave/cli/v3/command_run.go:294 +0x2416
github.com/urfave/cli/v3.(*Command).Run(...)
        /home/runner/work/cli/cli/vendor/github.com/urfave/cli/v3/command_run.go:93
main.main()
        /home/runner/work/cli/cli/scalingo/main.go:142 +0xb77
```

briceamen:
Fixes #1207.

This PR prevents `scalingo run` from panicking when operation polling fails after the one-off has been started. It adds a regression test for the polling error path and keeps the waiter on the last successful operation state instead of dereferencing a nil operation.

Tests:
- `go test ./apps -run TestWaitOperationReturnsPollingErrorWithoutPanic -count=1`
- `go test ./...`


apps/operation_waiter.go
EtienneM: issue: `operationsService` must be initialized here. Not in `WaitOperation`

apps/operation_waiter_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/Scalingo/go-scalingo/v11"
+)
+
+type testOperationsService struct {
+	show func(ctx context.Context, app, opID string) (*scalingo.Operation, error)
+}
+
+func (s testOperationsService) OperationsShow(ctx context.Context, app, opID string) (*scalingo.Operation, error) {
+	return s.show(ctx, app, opID)
+}
```
EtienneM: question: shouldn't we use a mock, like in 99% of our Go code?

EtienneM (changes requested):
question: I don't understand the fix, could you explain?

briceamen:
> question: I don't understand the fix, could you explain?

The panic came from our error handling, I was able to reproduce it using a TDD approach [`687223d` (this PR)](https://github.com/Scalingo/cli/pull/1211/commits/687223d4a9a4a52f9fc8bcaaec3cfa272b99b29b), not from the operation itself. A failed poll could return `nil, err,` and we  were then dereferencing `op.ID`. The fix removes that nil dereference and returns the polling error normally.

apps/operation_waiter_test.go
briceamen: I checked the current repo and `go-scalingo` but there is no generated `OperationsService` mock available it seems. I kept the test double small because this test only needs one `OperationsShow` path. If we want, I can add a generated mock in a follow-up

apps/operation_waiter.go
briceamen: Fixed. `operationsService` is now initialized when the waiter is created

apps/operation_waiter.go
```diff
@@ -20,23 +20,35 @@ const (
...
+	if err != nil {
+		return nil, errors.Wrap(ctx, err, "get Scalingo client")
 	}
+
+	return &OperationWaiter{
+		output:            output,
+		app:               app,
+		url:               url,
+		prompt:            defaultOperationWaiterPrompt,
+		operationsService: c,
```
matthieu526-scalingo, line 45: question: is `operationsService` the right name? Here this is really a scalingo client.

matthieu526-scalingo (approved):
SGTM, but I think the second review would still be nice.

apps/operation_waiter_test.go
EtienneM: But we have plenty of mocks in go-scalingo. Hence I think a correct fix is to generate the mock for `OperationsService` in go-scalingo and use it here.

apps/operation_waiter.go
```diff
@@ -9,10 +9,11 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/Scalingo/cli/config"
-	"github.com/Scalingo/cli/io"
 	"github.com/Scalingo/go-scalingo/v11"
 	"github.com/Scalingo/go-utils/errors/v3"
+
+	"github.com/Scalingo/cli/config"
+	"github.com/Scalingo/cli/io"
```
EtienneM: nitpick: remove extra new line

```suggestion
"github.com/Scalingo/go-scalingo/v11"
"github.com/Scalingo/go-utils/errors/v3"
	"github.com/Scalingo/cli/config"
	"github.com/Scalingo/cli/io"
```

apps/operation_waiter.go
```diff
@@ -24,19 +25,31 @@ type OperationWaiter struct {
 	prompt string
 	app    string
 	url    string
+	client scalingo.OperationsService
```
EtienneM: issue: as opposed to what Matthieu said, I think this should still be named `operationsService`. In this context we only care about the interface, not the actual implementation
```suggestion
	operationsService scalingo.OperationsService
```

apps/operation_waiter.go
```diff
@@ -24,19 +25,31 @@ type OperationWaiter struct {
...
+	return newOperationWaiter(ctx, os.Stderr, app, url)
 }
 
-func newOperationWaiter(output stdio.Writer, app, url string) *OperationWaiter {
+func newOperationWaiter(
+	ctx context.Context,
+	output stdio.Writer,
+	app,
+	url string,
+) (*OperationWaiter, error) {
```
EtienneM: suggestion: I find this way of displaying the arguments unreadable, I would prefer in a single line or max two lines if it's really too long. But I don't think it's the case here.
```suggestion
func newOperationWaiter(ctx context.Context, output stdio.Writer, app, url string) (*OperationWaiter, error) {
```

apps/operation_waiter.go
```diff
@@ -62,18 +70,19 @@ func (w *OperationWaiter) WaitOperation(ctx context.Context) (*scalingo.Operatio
...
-	op, err = c.OperationsShow(ctx, w.app, opID)
+	op, err = w.client.OperationsShow(ctx, w.app, opID)
 	if err != nil {
 		return nil, errors.Wrapf(ctx, err, "get operation %v", opID)
 	}
 
 	go func() {
 		for {
-			op, err = c.OperationsShow(ctx, w.app, opID)
+			nextOp, err := w.client.OperationsShow(ctx, w.app, opID)
```
EtienneM: suggestion: I'm not keen on the `nextOp` name. We usually name this kind of variable something like `reloadedThing` when we need to reload something to check its state
```suggestion
			reloadedOp, err := w.client.OperationsShow(ctx, w.app, opID)
```

EtienneM (changes requested):

vendor/github.com/Scalingo/go-scalingo/v11/scalingomock/addonprovidersservice_mock.go
EtienneM, line 1: issue: how can this be vendored without a new release of go-scalingo?

EtienneM (changes requested):

vendor/github.com/Scalingo/go-scalingo/v11/scalingomock/addonprovidersservice_mock.go
briceamen: The issue was more that the cli was using  `go-scalingo` v11 but the mocks were not vendored. They were here all along https://github.com/Scalingo/go-scalingo/blob/master/scalingomock/addonprovidersservice_mock.go

And I don't really know why 🤷

EtienneM (approved):

PR #1220, opened 2026-04-28
fix(addons): parse `maintenance-window-hour` as an int

### LINKED ISSUE #1219: [`addons-config `] `--maintenance-window-hour` always set the same hour, whatever the value
```
sscalingo --app REDACTED addons-config ad-REDACTED --maintenance-window-hour 3
Addon config updated.
```

But on the dashboard:

Image

With the debug output, we can see that the payload sent to the API is always the same.

Notify when fixed: https://www.notion.so/HULK-3604

EtienneM:
```
bsscalingo --app REDACTED addons-config ad-REDACTED --maintenance-window-hour 3
Addon config updated.
```

And now on the dashboard:

image

Fix #1219 

### DIFF
--- a/cmd/addons.go
+++ b/cmd/addons.go
@@ -199,21 +199,21 @@ var (
 
 			currentApp := detect.CurrentApp(ctx, c)
 			currentAddon := c.Args().First()
-			config := addons.UpdateAddonConfigOpts{}
+			updateAddonConfig := addons.UpdateAddonConfigOpts{}
 
 			if c.IsSet("maintenance-window-day") {
-				config.MaintenanceWindowDay = utils.StringPtr(c.String("maintenance-window-day"))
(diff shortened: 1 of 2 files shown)


cmd/addons.go
```diff
@@ -183,7 +183,7 @@ var (
 		Category: "Addons",
 		Usage:    "Configure an add-on attached to your app",
 		Flags: []cli.Flag{&appFlag,
-			&cli.StringFlag{Name: "maintenance-window-hour", Usage: "Configure add-on maintenance window starting hour (in your local timezone)"},
+			&cli.IntFlag{Name: "maintenance-window-hour", Usage: "Configure add-on maintenance window starting hour (in your local timezone)"},
```
EtienneM, line 186: This is the fix. The rest is just a variable renaming.

curzolapierre (approved):
Thanks for the fix 🙏 

LGTM

PR #1229, opened 2026-06-10
feat(dbdr/net-peering) Add commands to configure net peerings of DB-DR

leo-scalingo:
* `database-network-configuration`: Get networking information on a database to create a net peering
* `database-net-peerings`: List all net peerings
* `database-net-peerings-add`: Create a net peering to the database based on an outscale net peering request
* `database-net-peerings-remove`: Delete a netpeering
* `database-endpoints`: List the different types of endpoints of a given database (public/private rw/ro access)

- [x] Add a changelog entry in the section "To Be Released" of CHANGELOG.md

PR #1234, opened 2026-06-12
deps(urfave/cli) deps(go/crypto) Bump versions which were downgraded in #1229 without reason

leo-scalingo:
See #1229


go.mod
```diff
@@ -23,9 +23,9 @@ require (
 	github.com/sirupsen/logrus v1.9.4
 	github.com/stretchr/testify v1.11.1
 	github.com/stvp/rollbar v0.5.1
-	github.com/urfave/cli/v3 v3.8.0
+	github.com/urfave/cli/v3 v3.9.0
 	go.uber.org/mock v0.6.0
-	golang.org/x/crypto v0.50.0
+	golang.org/x/crypto v0.43.0
```
EtienneM: I don't understand why the versions of `github.com/go-git/go-git/v5` and `golang.org/x/crypto` are downgraded

EtienneM (changes requested):

go.mod
leo-scalingo: Woupsy sorry, `go get -u` did some mess here

leo-scalingo: Fixed

EtienneM (approved):

PR #1239, opened 2026-07-10
build(workflows): pin all actions versions

EtienneM:
Not relevant:


matthieu526-scalingo (approved):
🧹 
Does it still require a changelog entry?

EtienneM:
> 🧹 Does it still require a changelog entry?

I don't think so. That's why I didn't add one

PR #1244, opened 2026-08-05
feat(pitr): add command to restore a database to a specific point in time

matthieu526-scalingo:
Add a new command to restore PITR.
It reuses the same mechanism as maintenance. Both components are shared between SR addons and DR databases.
It adds a small check to ensure the time is correctly parsed.

- [x] right now depending on this [PR](https://github.com/Scalingo/go-scalingo/pull/512). Before merging it should be set to a released version (or not?)
- [X] QA not executed yet
  - Working as expected, restoration is triggered

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,9 @@ COMMANDS:
    Notifiers - Global:
      notification-platforms  List all notification platforms
 
+   PITR management:
+     database-pitr-restore  Restore a database to a specific point in time
+
    Private Networks:
      private-networks-domain-names  List the private network domain names of an application
 
@@ -243,6 +246,7 @@ COMMANDS:
 GLOBAL OPTIONS:
    --addon string              ID of the current addon (default: "<addon_id>") [$SCALINGO_ADDON]
    --app string, -a string     Name of the app (default: "<name>") [$SCALINGO_APP]
+   --format string             [json|table] (default: "table")
    --remote string, -r string  Name of the remote (default: "scalingo")
    --region string             Name of the region to use
    --help, -h                  show help
--- a/db/pitr/restore.go
+++ b/db/pitr/restore.go
@@ -2,6 +2,7 @@ package pitr
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/Scalingo/cli/config"
@@ -13,6 +14,15 @@ func Restore(ctx context.Context, currentResource, addonName string, restoreTime
 	c, err := config.ScalingoClient(ctx)
 	if err != nil {
 		return errors.Wrapf(ctx, err, "get Scalingo client")
+	}
+
+	io.Warning("This operation is irreversible and previous data will be restore")
+	io.Info("Do you want to confirm? (y/n)")
+
+	var confirm string
+	_, _ = fmt.Scanln(&confirm)
+	if confirm != "y" && confirm != "Y" {
+		return errors.New(ctx, "You didn't confirm, aborting…")
 	}
 
 	operationID, err := c.DatabaseRestorePITR(ctx, currentResource, addonName, restoreTime)
--- a/db/pitr/restore.go
+++ b/db/pitr/restore.go
@@ -1,0 +1,26 @@
+package pitr
+
+import (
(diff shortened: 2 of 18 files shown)


db/pitr/restore.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/Scalingo/go-utils/errors/v3"
+)
+
+func Restore(ctx context.Context, currentResource, addonName string, restoreTime time.Time) error {
+	c, err := config.ScalingoClient(ctx)
+	if err != nil {
+		return errors.Wrapf(ctx, err, "get Scalingo client")
+	}
+
+	operationID, err := c.DatabasePITRRestore(ctx, currentResource, addonName, restoreTime)
```
curzolapierre: **Suggestion**: what about adding a confirmation message to the end user?
Which would format the date as human readable (maybe add the local time conversion).

README.md
curzolapierre, line 1: **Praise** thanks for this update

curzolapierre (commented):
Just a question otherwise LGTM

db/pitr/restore.go
matthieu526-scalingo: What do you mean? There is a `io.Status` 2 lines below. Or do you mean improving said message?

db/pitr/restore.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		return errors.Wrapf(ctx, err, "get Scalingo client")
+	}
+
+	operationID, err := c.DatabaseRestorePITR(ctx, currentResource, addonName, restoreTime)
+	if err != nil {
+		return errors.Wrap(ctx, err, "restore database")
+	}
+
+	io.Statusf("Database restore operation %s has been created", operationID)
```
SCedricThomas: question: what does the output looks like?

matthieu526-scalingo: I am not sure what you mean? It will just display a line with that info, like we do on many other commands (eg project creation)

SCedricThomas: I am asking because there is no "\n" at the end. Usually I would expect a newline at the end

matthieu526-scalingo: Oooooh. I see what you mean, nice catch :bow: . I guess the example I used was wrong.
I haven't QAed yet (as stated in the description)

db/pitr/restore.go
curzolapierre: Sorry it was not clear: I had in mind something like apps destroy command, ask for an user confirmation before to proceed the operation:
https://github.com/Scalingo/cli/blob/93a19478b506e809b1dd440d6ab09cc530d51ae6/apps/destroy.go#L29

matthieu526-scalingo: Good question! Usually I'd have asked Benjamin, or checked how it is in the dashboard. But you are right, as it is a destructive action it might be better.

matthieu526-scalingo: Added here fc07c8c5613b38dabe4177b1cac2f7ee7e63fc87. I added a yes/no more than a name, as the use can use the addon name or ID or type, it didn't see great.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.25.0
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.7
-	github.com/Scalingo/go-scalingo/v11 v11.1.1
+	github.com/Scalingo/go-scalingo/v11 v11.1.2-0.20260810154739-f30bd2e7ef2c
```
SCedricThomas: reminder: don't forget to release v11.1.2 before merging this PR

SCedricThomas (approved):
LGTM, just a reminder. Ping me when you need a reapproval post release

curzolapierre (approved):

go.mod
matthieu526-scalingo: (:information_source: discussion about that [here](https://scalingo.slack.com/archives/C01E42AT2G6/p1786454269005829))

matthieu526-scalingo: Done.

PR #1248, opened 2026-08-10
task(deps): update go-scalingo to remove support for the db-api

### LINKED ISSUE #1247: Remove support for internal db-api
The db-api has been de-provisioned, and a redirection is used. But it means we can remove the support and point to the API only.
It should be done after https://github.com/Scalingo/go-scalingo/issues/514 is done.

matthieu526-scalingo:
The DB API has been deprovisionned and is not required anymore.
This is pending confirmation from UFS [here](https://scalingo.slack.com/archives/C6GE4JP3L/p1786358646421339).

Based on [this](https://github.com/Scalingo/go-scalingo/pull/515) go-scalingo PR.

- [ ] Update go-scalingo to the new released version

Fix #1247 .

### DIFF
--- a/cmd/firewall_rules.go
+++ b/cmd/firewall_rules.go
@@ -41,7 +41,7 @@ var (
 
 			utils.CheckForConsent(ctx, databaseName, utils.ConsentTypeDBs)
 
-			err = dbng.FirewallRulesList(ctx, databaseName, addonID)
+			err = dbng.FirewallRulesList(ctx, addonID)
 			if err != nil {
 				errorQuit(ctx, err)
 			}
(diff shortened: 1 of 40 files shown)


cmd/firewall_rules.go
```diff
@@ -193,32 +193,18 @@ var (
...
+		Name:     "database-firewall-managed-ranges",
+		Category: "Databases DR",
+		Usage:    "List available managed ranges",
 		Description: CommandDescription{
-			Description: "List all available managed ranges for firewall rules of a database",
+			Description: "List all available managed ranges for firewall rules",
 			Examples: []string{
-				"scalingo database-firewall-managed-ranges my-db-id",
-				"scalingo --database my-db database-firewall-managed-ranges",
+				"scalingo database-firewall-managed-ranges",
```
sc-david-voisin, line 202: praise: good catch, they are not linked to a database

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.25.0
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.7
-	github.com/Scalingo/go-scalingo/v11 v11.1.1
+	github.com/Scalingo/go-scalingo/v11 v11.1.2-0.20260810110303-774d8583f620
```
sc-david-voisin, line 7: question: the incoming `go-scalingo` CHANGELOG indicated "breaking change", so I suppose this version is not the definitive one ?

sc-david-voisin (commented):
SGTM, one question though

go.mod
matthieu526-scalingo: Yes, that's exactly why I put ` Update go-scalingo to the new released version` in the description :slightly_smiling_face: 
In any case I'll be waiting on UFS before going further.

sc-david-voisin: Ok, I misunderstood its meaning.
And if you wait for UFS, it is all good!

sc-david-voisin (approved):

matthieu526-scalingo:
Closing this for now. The behaviour is not the one expected. The switch of endpoint comes with a side effect, where the "prefix" are different. IE `/v1` vs `/api`. It would require keeping both clients, or refactor in a more complex way. This is not worth it for now as this is a temporary situation.

